PR #3, opened 2025-10-14
Dns check and transparent background opentui patch

KeremP:
Add DNS ownership check by querying DNS records of hostname domain parsed from target url. Skips if local.


joshkotrous:
@KeremP  cleared conflicts, however did you finish building this out? When entering a url i get the below immediately. Can you please add an alert dialog to display the text record to veirfy and handle the verification itself? 

Screenshot 2025-10-29 at 1 49 06 PM

joshkotrous (changes requested):
see pr comment

Issue #39, opened 2025-11-20
TypeError When Viewing Previous Sessions


kylejryan:
If you go to open/view a session by pressing O, it'll output a TypeError. See attached. 

<img width="875" height="331" alt="Image" src="https://github.com/user-attachments/assets/ad9bfc6b-aad2-4af9-90d0-885a3286d83d" />

kylejryan:
Agent works as expected otherwise.

kylejryan:
Resolved in PR #45. Deleted source branch after merge and closing issue.

PR #45, opened 2025-11-23
Fix session type error

KeremP:
Fixes type error when viewing old sessions that occurs when messages are parsed - using zod to parse and ensure type safety.

Addresses #39.


KeremP:
@kylejryan if you want to test this quick

kylejryan (approved):
LGTM! I ran a smoke test and couldn't replicate the bug

PR #49, opened 2025-11-29
Session & Storage management updates + overall design updates

KeremP:
This is somewhat large and currently WIP

Implemented Session and Storage management w/ locks to ensure safe read/writes to local disk and improving DX around session management.

Results are saved to local json-based "database" "tables" for cleaner serialization.


KeremP:
Merged in canary and fixed a ton of bugs.

We now have:
- swarm dashboard (dash view of pentest sub agents)
- meta prompting agent (uses external memory file to track attempts and pivot methodology if not working, rewrite its execution prompt, and is spawned with specific exploit objectives determined by the attack surface agent)
- updated daytona benchmark script to spawn validation benchmarks in parallel
- improved persistence to disk with lock management
- improved ui design

I also removed a few scripts that were incompatible with the current design. we can add those back in as we update the "views"

joshkotrous:
Need to update docs in console repo to reflect new commands

KeremP:
implemented `driver` mode and option flags to `/web` command (replacing `/init`)

bug fixes for benchmarks

bug fixes for some tui things

PR #65, opened 2025-12-17
added message streaming

kylejryan:
Added streaming in varying ms to agent outputs.


KeremP:
This is great! Just one thing - it's streaming just the tool calls but not the messages themselves

image

kylejryan:
@KeremP Take a peek now. I got the messages to appear but they're giving some trouble streaming the characters out (minor UX item so maybe that's an issue for another day)

PR #67, opened 2025-12-30
Resolved Pressing Esc To Close Shortcut Menu Ending Pentest

kylejryan:
## Problem

When pressing `?` to view keyboard shortcuts while in a `/web` pentest session (auto or manual mode), pressing Escape to close the dialog would also exit the entire pentest back to the home screen.

## Root Cause

In this TUI framework, `key.preventDefault()` does not stop other `useKeyboard` handlers from firing—each handler receives the keypress independently. The ShortcutsDialog used local state (`showShortcutsDialog`) but child components (SwarmDashboard/DriverDashboard) had no way to know a dialog was open, so they also processed the escape key.

## Solution

Extended the existing DialogContext to track external dialogs:

1. Added `externalDialogOpen` boolean to DialogContext
2. Set it when ShortcutsDialog opens/closes
3. SwarmDashboard and DriverDashboard now check this flag before processing keyboard events

This approach:

- **Minimal**: Extends existing infrastructure rather than creating new patterns
- **Correct**: React's batched updates guarantee the flag check works regardless of handler execution order
- **Efficient**: O(1) boolean check in keyboard handlers

## Files Changed

- `src/tui/components/dialog.tsx` - Added `externalDialogOpen` state to context
- `src/tui/index.tsx` - Track dialog state when opening/closing ShortcutsDialog
- `src/tui/components/swarm-dashboard/index.tsx` - Skip keyboard handling when dialog open
- `src/tui/components/driver-dashboard/index.tsx` - Skip keyboard handling when dialog open
- `src/tui/components/commands/shortcuts-dialog.tsx` - Removed unnecessary `preventDefault()`

## Testing

1. Start `/web` pentest (auto or manual mode)
2. Press `?` to open shortcuts dialog
3. Press Escape to close dialog → pentest view remains active ✓
4. Press Escape again → exits to home as expected ✓

### DIFF
--- a/src/tui/components/dialog.tsx
+++ b/src/tui/components/dialog.tsx
@@ -54,6 +54,8 @@ interface DialogContextValue {
   stack: DialogStackItem[];
   size: "medium" | "large";
   setSize: (size: "medium" | "large") => void;
+  externalDialogOpen: boolean;
+  setExternalDialogOpen: (open: boolean) => void;
 }
 
 const DialogContext = createContext<DialogContextValue | null>(null);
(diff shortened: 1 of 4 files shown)


KeremP (approved):
looks good - thanks for this!

eventually we should probably clean up how we do key binding registration/handling lol

PR #68, opened 2025-12-30
Stop Agent Execution, HTTP Requests, and Commands When Exiting Pentest Session

kylejryan:
## Summary

- Fix abort signal not being triggered when pressing ESC to exit a pentest
- Stop LLM calls and token consumption immediately on exit
- Propagate abort signal to all pentest tools so HTTP requests and commands stop immediately

## Problem

When pressing ESC to exit a `/web` pentest session:

1. LLM calls continued—token count kept climbing after exit
2. HTTP requests kept hitting the target server
3. Commands continued executing in the background

Root cause: Two issues:

1. `handleBack()` in session-view was not calling `abort()` on the AbortController—so the AI SDK never received the abort signal
2. Tools (`http_request`, `execute_command`, `fuzz_endpoint`) created independent AbortControllers and never received the parent abort signal

### 1. Fix abort trigger in session-view (`session-view/index.tsx`)

- Actually call `abortController.abort()` in `handleBack()` when ESC is pressed
- Add cleanup effect to abort on component unmount
- This stops the AI SDK stream, halting LLM calls and token consumption

### 2. Propagate abort signal to tools (`tools.ts`)

- Add `abortSignal?: AbortSignal` parameter to `createPentestTools()`
- `http_request`: Use `AbortSignal.any()` to combine parent abort + timeout signals
- `execute_command`: Check abort signal before/after execution
- `fuzz_endpoint`: Pass abort signal to fetch calls, check before each request

### 3. Pass abort signal from all agents

- `attackSurfaceAgent/agent.ts`
- `metaTestingAgent/agent.ts`
- `metaVulnerabilityTestAgent.ts`

## Test Plan

1. Start `/web` pentest against a target
2. Wait for HTTP requests to begin (visible in target logs or network)
3. Press ESC to exit the pentest
4. Verify token count stops increasing immediately
5. Verify no new LLM calls are made
6. Verify no new HTTP requests hit the target server
7. Verify graceful error messages ("Request aborted by user") in logs

### DIFF
--- a/src/core/agent/attackSurfaceAgent/agent.ts
+++ b/src/core/agent/attackSurfaceAgent/agent.ts
@@ -95,7 +95,8 @@ export async function runAgent(opts: RunAgentProps): Promise<{
     session,
     model,
     toolOverride,
-    onToolTokenUsage
+    onToolTokenUsage,
+    abortSignal
   );
 
   // Attack Surface specific tool: document_asset
--- a/src/core/agent/metaTestingAgent/agent.ts
+++ b/src/core/agent/metaTestingAgent/agent.ts
@@ -158,7 +158,9 @@ export async function runMetaTestingAgent(
   const baseTools = createPentestTools(
     session,
     model,
-    toolOverride
+    toolOverride,
+    undefined,  // onTokenUsage - not needed here
+    abortSignal
   );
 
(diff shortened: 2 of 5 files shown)


KeremP (approved):
nice, ty sir

PR #76, opened 2026-01-14
Add objective and evidence criteria guardrails and remove execution prompt iteration for meta agents

KeremP:
add evidence criteria, add objective guard rails, remove execution prompt iteration, improve shared memory between agents, llm-based orchestrator


joshkotrous:
Tested on https://github.com/pensarai/console/pull/548 and it doesnt seem to be persistent enough still on coffee-shop. Its continuing onto the next objective in the plan without thoroughly testing each one hence still missing the sql injection.

Perhaps we need to split each objectvie into its own agent.

It also opted to not create a poc for some reason when it found an ssrf because it thought its manual verification was enough? 

```
- Let me document the SSRF finding without POC since I've already confirmed the vulnerability through manual testing:

- Called document_finding tool:
metadata
{
  "error": "POC_NOT_FOUND",
  "message": "POC not found at: pocs/ssrf_poc_placeholder.txt\n\nCreate POC first using create_poc tool, then call document_finding.",
  "success": false
}

```

KeremP:
hm looking into this

PR #94, opened 2026-01-30
Implemented /pentest and /operator

joshkotrous:
Wired /pentest to the existing /web command with --swarm option to automatically launch the swarm based pentest as a default 

Added /operator command wired to the operator mode


joshkotrous:
Still a minor bug with some tool updates in the Attack Surface Agent but this is ready

PR #123, opened 2026-02-11
Show hint message for non-command input on home screen

ChinmayShringi:
## Summary
- When users type normal text (not a `/command`) on the home screen and press Enter, the app previously crashed by calling `onStartSession()` which attempted to create an exploration session
- Now shows a friendly **"Type /help to get started"** hint message in cream text that auto-clears after 3 seconds
- Input is cleared on submit, and `/commands` continue working as before

## Changes
- **`src/tui/components/chat/home-view.tsx`**: Replaced `onStartSession(value)` call in `handleSubmit` with a hint message via `useState`/`useEffect`

## Test plan
- [x] On the home screen, type normal text (e.g. "Hi") and press Enter → hint appears, app does not crash
- [x] Hint auto-disappears after ~3 seconds
- [x] Typing `/help` still works as expected
- [x] Other `/commands` with autocomplete still function normally

### DIFF
--- a/src/tui/components/chat/home-view.tsx
+++ b/src/tui/components/chat/home-view.tsx
@@ -7,7 +7,7 @@
  * - Centered command input with inline autocomplete
  */
 
-import { useCallback } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { RGBA } from "@opentui/core";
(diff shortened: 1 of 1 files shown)


KeremP (approved):
Nice! Thanks

PR #124, opened 2026-02-11
Fix light mode text visibility (#120)

### LINKED ISSUE #120: Light mode text is not visible

ChinmayShringi:
## Summary

- Auto-detect terminal background (light/dark) at startup using `COLORFGBG` env var and macOS `AppleInterfaceStyle`
- Add centralized theme system with `ThemeProvider` context and `useColors()` hook
- Define dark and light color palettes with semantic color names (`primaryText`, `secondaryText`, `overlayBg`, etc.)
- Migrate 56 files from hardcoded `RGBA.fromInts()` constants and string literals (`fg="white"`, `fg="gray"`) to use theme-aware colors

## Test plan

- [x] Run with dark terminal — appearance should be identical to current
- [x] Run with a light terminal profile: all text should be readable
- [x] Check critical screens: disclosure, home, chat messages, footer, help dialog, config, operator wizard
- [x] Verify `npx tsc --noEmit` passes cleanly

Closes #120


joshkotrous:
Hey @ChinmayShringi! Thanks for opening this PR 🎉 however I just tested with your branch and it appears input text is still not displaying in light mode: 

https://github.com/user-attachments/assets/f599fa27-3103-4c3e-b201-b5be2a7cd57f


Did you experience different behavior? What terminal and OS are you running on?

ChinmayShringi:
Hello @joshkotrous,
I tested it on zsh(Z shell). Could you tell me yours I will test on that.

https://github.com/user-attachments/assets/d9168c4f-f06c-4016-8057-8d33f6ee919f

joshkotrous:
@ChinmayShringi I tested on Mac OS also using zsh

joshkotrous:
@jorgeraad  do you think you could pls look at this and see if you can replicate?

jorgeraad:
First off, @ChinmayShringi for contributing. Really appreciate you opening a PR

I did test on my setup (also zsh on mac) and I'm seeing the same issue.

- top left: before change, light background in terminal
- top right: before change, dark background in terminal
- bottom left: after change, light background in terminal
- bottom right: after change, dark background

light-mode

Aside from that though, I'm thinking we may not want to change the text color solely based on system light/dark mode unless we also set a contrasting background in the TUI. Users can often have terminal setups that don't match their system theme. For example, I switch between light/dark themes but always have a dark background in my terminal.

If we proceed with this PR, we should:
- set the TUI background color to match the system theme (questionable as users may have set their terminal theme to match their preference, so we probably shouldn't ignore and force them to change their system theme)
- create a theme system and allow the user to switch between themes in the TUI. We could start with simple light and dark themes.

@joshkotrous lmk what you think.

jorgeraad:
Actually @ChinmayShringi, reread your PR summary and realized you intended to change the text color based on the terminal background, not the system background, which makes sense. but it seems it actually changes with the system theme.

For example, here the top was created in dark mode and bottom in light mode:
Screenshot 2026-02-17 at 10 14 28

jorgeraad:
Hey @ChinmayShringi, your issue prompted us to [introduce themes](https://github.com/pensarai/apex/pull/161)! Since this should handle the light mode problem, I'm going to close this PR. Thanks for opening this and let us know if you run into any more issues!

PR #127, opened 2026-02-12
Initial implementation of general offensive security agent harness (#128)

joshkotrous:
This PR implements a general `OffensiveSecurityAgent` harness in the form of a class. This class can be extended to create new agent definitions purely as configuration. 

Added `/api` for centralized entry point to run individual specialized agents.

See `/src/core/offensiveAgent` for the general agent definition

See `/src/core/agents` for specialized agent definitions

Existing agents have been moved to `/agents/legacy`

Not married to this implementation but opened as a starting point since I was thinking about it

Addresses: https://github.com/pensarai/apex/issues/128

**Updates**
- Wired updated agent harness to `/pentest` and `/operator`
- Added whitebox attack surface analysis agent
- Added whitebox pentest support that uses whitebox attack surface analysis + pentest agent swarm
- Added headless commands to run `pensar pentest` and `pensar target-pentest`

Yes some things you might be expecting may have been removed, but the intention is to strip down as much as possible so we can build off a better foundation with what weve learned.


joshkotrous:
@kylejryan to review auth agent implementation

@Yuvanesh-ux to review for benchmark performance

@jorgeraad to review for visibility

jorgeraad:
LGTM. Really strong cleanup that makes it way easier for me as a new contributor to navigate the codebase and makes it significantly easier to work with.

I have some small questions/suggestions, but most of this is just general observations and future thinking given that this PR is a major refactor.

Since I think we should merge this asap so we can build on top of it and most are not specific to this PR, I decided to just put it all here instead of scattering it throughout the code base.

<details> 

<summary>Click to expand</summary>

### Minor notes

- `generateReport` and `ai.test.ts` look like they have commented-out code. Just want to confirm we need these or if we can delete them outright?
- `codeAgent` is a bit misleading since it reads/analyzes code but doesn't write it. Something like `codeAnalysisAgent` would be clearer.
- A few agent prompts (code agent, whitebox agent) re-describe information that's already in the tool definitions. We should try to keep the contextual instructions specific to the task and only add to prompts what the agent wouldn't get from the tool definitions alone.
- The dynamic import in `delegateAuth.ts` to break the circular dependency is minor, but yeah we should clean it up at some point.

### Related to this PR: callbacks => event bus

The callback system introduced here for the agents (e.g., `onTextDelta`, `onToolCall`) is one of the main parts of this refactor that makes a huge difference. I'd say the natural next evolution is to take it one step further and implement a proper event bus. It would clean up the forwarding and make it easier to go fully event-driven. This refactor is already a huge improvement so this isn't necessarily critical, but yeah this will also make it a lot easier to implement as a follow-up.

### Unrelated to this PR:

- There's a lot of direct `writeFileSync` throughout the codebase. We should introduce a more cohesive data access layer that abstracts persistence. As a plus, having that abstraction could potentially let us swap the persistence layer depending on the client if that's something we'd want at some point.
- There's a good amount of structural overlap between the specialized agents' input/output types. Extracting some common interfaces could reduce duplication and make it easier to move closer towards teh goal of more flexible, composable agents.

### Longer-term thoughts/open questions (kind of all related)

#### Fewer, simpler tools

My intuition says some of the specialized tools might be doing more than they need to. The agent already has strong primitives (`execute_command`, `http_request`, browser tools), and I'm curious whether some of the higher-level tools could be replaced by better prompting around those primitives.

For example, we have both `delegate_to_auth_subagent` and `authenticate_session`. Could having both make it harder for the agent to figure out which one to reach for? Or could these be collapsed into a single prompt-guided flow, describe the auth approach in natural language and guide the agent through it with the simpler tools it already has? If we can work towards this and prove this, the result would be a more flexible, creative, and resilient agents.

Something worth exploring is how much we lean into specialized tools versus just letting the agent write bash commands. Those are well-represented in training data and give the agent a lot of flexibility to experiment and retry when things don't work.

Fortunately, now that this refactor has cleanly separated the specialized agents and their tools, we can experiment with this gradually. We can try replacing individual tools or specialized agents with a more general approach where the agent works with fewer tools and is guided with more natural language instead.

#### Less rigid inter-stage types

The typed inputs/outputs between stages make it easier to get something working (e.g., it forces the model to produce complete structured data, and helps the agent follow instructions, it adds guardrails, it makes it easier to display things in the UI). That said, especially in cases where we're piping the output of one agent directly as input to the next, do we need to force a rigid schema there? The schema can limit the agent's creativity and make it hard to recover from partial results. For example, the blackbox agent's asset schema asks the agent to fill specific expected fields one by one, but as an example, what if the most interesting parameters for a given target don't fit those categories?

I wonder if there's a middle ground. Maybe looser key-value passing between stages with a validation step, so agents can surface things we didn't anticipate while still maintaining enough structure. Could be worth experimenting with.

#### Config-driven vs code-driven agents

I'm wondering what it'd look like if we move toward defining agent specializations via configuration rather than requiring a new code file for each one. With the subagent design introduced this PR, it already feels like we're kind of already doing that, just in code. Our specialized agents are already just prompts + toolset + guidance on input and output. What if this can be defined in configuration, allowing for easy custimization. Agents could learn and create new configurations based on learning. Users can add configurations specific to their targets instead of being locked into our implementations. And less rigid code, could lead to more flexible, creative, and resilient agents, as mentioned above. And most importantly, as the models get better, they are not limited by our harness.

These 3 points kind of all fit together and fall in line with the pattern we're seeing take off with agents like Claude Code:

- spawn subagents generously
- define "skills" (iow specialized prompts specific to a task) that can be handed off to these subagents
- fewer tools (and more bash?)

As we move in this direction, we can write evals for specific existing workflows/specialized agents and make gradual changes, checking for regressions as we go along. This refactor will make it way easier iterate on this.

---

LGTM. Let's merge asap!


</details>

joshkotrous:
@kylejryan  just to guide your review of the auth and save you some time - I think the main place I may have broken something was specific to the auth persistence of the found cookies / token from the browser. I reimplemented this in a slightly different way than you had with the `AuthStateManager`.

joshkotrous:
@jorgeraad 


<details>

<summary>Click to expand</summary>

> * `generateReport` and `ai.test.ts` look like they have commented-out code. Just want to confirm we need these or if we can delete them outright?

Generate report is kept as context to bring back later, ai.test.ts the commented out tests relied on files that were deleted at some point. Just commented that out to move forward but well bring it back. I think its specific to testing context exhaustion errors which relied on a file to read from to exhaust the context. 

> * `codeAgent` is a bit misleading since it reads/analyzes code but doesn't write it. Something like `codeAnalysisAgent` would be clearer.

Well probably add write file to this eventually so we can implement patching in Apex


> ### Related to this PR: callbacks => event bus
> The callback system introduced here for the agents (e.g., `onTextDelta`, `onToolCall`) is one of the main parts of this refactor that makes a huge difference. I'd say the natural next evolution is to take it one step further and implement a proper event bus. It would clean up the forwarding and make it easier to go fully event-driven. This refactor is already a huge improvement so this isn't necessarily critical, but yeah this will also make it a lot easier to implement as a follow-up.
> 

Great to hear - these callbacks are also essential for integration in console. We want the proper strcutred return from the agent API function but also want access to the token by token stream for parsing agent narration. This is how we achieve that. 

> ### Unrelated to this PR:
> * There's a lot of direct `writeFileSync` throughout the codebase. We should introduce a more cohesive data access layer that abstracts persistence. As a plus, having that abstraction could potentially let us swap the persistence layer depending on the client if that's something 
we'd want at some point.

Not sure about this one, we basically have a single plane of data per agent execution. Everything goes into the sessions folder using the `core/sessions` module. Otherwise app level config is handled by the `core/config` module.

> * There's a good amount of structural overlap between the specialized agents' input/output types. Extracting some common interfaces could reduce duplication and make it easier to move closer towards teh goal of more flexible, composable agents.
> 

Agreed but were constantly changing these as we see fit

> ### Longer-term thoughts/open questions (kind of all related)
> #### Fewer, simpler tools
> My intuition says some of the specialized tools might be doing more than they need to. The agent already has strong primitives (`execute_command`, `http_request`, browser tools), and I'm curious whether some of the higher-level tools could be replaced by better prompting around those primitives.
> 

Agree but also disagree. I started here - ripping out tools etc. but you very quckly find that for our use case to get the data in the shape we need, utilizing the tools to coerce the agent to do that becomes necessary. Http request and some of these redundant ones can be removed / consolidated with minimal impact, but like for the auth agent i've found it does need the specifc tools to reliably export session data in the shape and location we expect. The reason for these tools with the exception of the redundant ones is to ensure reliability at scale. We need the agent to reason but in some areas (especially for integration with console) we need consistency in data structure and location. This ties into below.

> For example, we have both `delegate_to_auth_subagent` and `authenticate_session`. Could having both make it harder for the agent to figure out which one to reach for? Or could these be collapsed into a single prompt-guided flow, describe the auth approach in natural language and guide the agent through it with the simpler tools it already has? If we can work towards this and prove this, the result would be a more flexible, creative, and resilient agents.
> 

Those tools specifically are probably redundnant, but we do want to have auth as a subagent to ensure reliability + also to manage context budget. Having an individual agent for some of these steps ensures we have fresh context budget so they can complete their task reliably without impacting the context of its parent agent.

> Something worth exploring is how much we lean into specialized tools versus just letting the agent write bash commands. Those are well-represented in training data and give the agent a lot of flexibility to experiment and retry when things don't work.
> 

Agreed but covered previously our use case for specialized tools / subagents. Most specialized tools like `smart_enumerate` etc. Im pretty sure I removed or at least turned off in the agent config. 


> #### Less rigid inter-stage types
> The typed inputs/outputs between stages make it easier to get something working (e.g., it forces the model to produce complete structured data, and helps the agent follow instructions, it adds guardrails, it makes it easier to display things in the UI). That said, especially in cases where we're piping the output of one agent directly as input to the next, do we need to force a rigid schema there? The schema can limit the agent's creativity and make it hard to recover from partial results. For example, the blackbox agent's asset schema asks the agent to fill specific expected fields one by one, but as an example, what if the most interesting parameters for a given target don't fit those categories?
> 
> I wonder if there's a middle ground. Maybe looser key-value passing between stages with a validation step, so agents can surface things we didn't anticipate while still maintaining enough structure. Could be worth experimenting with.
> 

We do need the specific inputs / outputs between stages. Required for console + getting the data we need reliably. But open to where we can improve to allow flexibilty. Basically the base structure isnt negotiable as we rely on it in console, but if they agent wanted to add additional fields it could. 

> #### Config-driven vs code-driven agents
> I'm wondering what it'd look like if we move toward defining agent specializations via configuration rather than requiring a new code file for each one. With the subagent design introduced this PR, it already feels like we're kind of already doing that, just in code. Our specialized agents are already just prompts + toolset + guidance on input and output. What if this can be defined in configuration, allowing for easy custimization. Agents could learn and create new configurations based on learning. Users can add configurations specific to their targets instead of being locked into our implementations. And less rigid code, could lead to more flexible, creative, and resilient agents, as mentioned above. And most importantly, as the models get better, they are not limited by our harness.
> 

I wanted to have config driven agents but we also have a few requirements that make the class helpful: 
1. Dynamically building the prompt at runtime based off conditions
2. The ability to customize the return output of the agent itself by utilizing the `resolveResult` function defined in the class. 

> These 3 points kind of all fit together and fall in line with the pattern we're seeing take off with agents like Claude Code:
> 
> * spawn subagents generously
> * define "skills" (iow specialized prompts specific to a task) that can be handed off to these subagents
> * fewer tools (and more bash?)
> 

The lens im looking at apex through right now in terms of UX is we need to get people using it and get feedback. I want to go into this with strong opinions for how it should be used and let people tell us how to iterate from there. That is how we arrived at the `/pentest` mode (entrypoint, new user, 100% autonomous and very opinionated workflow) and the `/operator` which allows the user to drive and for more advanced pentesters. Definitely can extend this in the future as we see how its used in practice. 

> As we move in this direction, we can write evals for specific existing workflows/specialized agents and make gradual changes, checking for regressions as we go along. This refactor will make it way easier iterate on this.
> 

Yes were looking into ways to integrate a small suite of benchmarks into the CI 

</details>

Yuvanesh-ux:
LGTM!

# Benchmark Comparison: Consolidated (Feb 2-13) vs Feb 20 Re-runs

> **Consolidated**: Sonnet for 002-010, Haiku for chains | **Feb 20**: All Haiku
> Chain benchmarks 048-050 have different content in updated suite -- comparison is approximate.

## Single-Vuln Benchmarks

| Bench | Vuln Type | Recall (Old) | Recall (New) | Precision (Old) | Precision (New) | Findings (Old) | Findings (New) | Time (New) |
|-------|-----------|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| 002 | GraphQL Injection | 100% | 75% * | 33% | 16% | 3 | 19 | 5.5m |
| 003 | JWT Algo Confusion | 100% | 100% | 9% | 8% | 11 | 12 | 9.4m |
| 005 | Blind SQLi | 100% | 100% | 17% | 6% | 1 | 17 | 6.8m |
| 006 | Race Condition | 0% | 0% | -- | -- | 5 | 11 | 7.0m |
| 007 | Proto Pollution RCE | 100% | 100% | 11% | 6% | 9 | 17 | 7.3m |
| 008 | OAuth Bypass | 100% | 100% | 20% | 19% | 10 | 16 | 8.5m |
| 010 | SSRF Chain | 100% | 100% | 10% | 11% | 10 | 9 | 6.7m |

\* APEX-002 expected vulns increased from 1 to 4 in updated suite.

## Chain Benchmarks

| Bench | Name (Old / New) | Recall (Old) | Recall (New) | Prec (Old) | Prec (New) | Time (Old) | Time (New) |
|-------|-------------------|:---:|:---:|:---:|:---:|:---:|:---:|
| 045 | IoT Compromise / IoT Command Center | 50% | **75%** | 7% | 13% | 24.0m | 34.7m |
| 046 | SSO Bypass / SAML Bypass | CRASH | **20%** | -- | 9% | -- | 8.9m |
| 048 | Service Mesh / LegacyBridge | 50% | **60%** | 14% | 16% | 19.9m | 36.2m |
| 049 | GameServer RCE / WAF SQLi | 67% | 63% | 50% | 39% | 14.0m | 7.1m |
| 050 | Supply Chain / WAF XSS | 83% | 40% | 36% | 18% | 14.8m | 6.1m |


## Takeaway

- **Recall is stable on single-vuln targets**: 5/7 held at 100%, race condition (006) still missed on both Sonnet and Haiku
- **Haiku is noisier**: ~2x more findings per run (avg 14 vs 7), tanking precision without improving recall
- **Chains 045/046/048 improved**: +25%, +20% (was crash), +10% recall respectively
- **049/050 can't be compared**: Benchmark content completely changed (GameServer->WAF SQLi, Supply Chain->WAF XSS)

Yuvanesh-ux (approved):

jorgeraad (approved):

kylejryan:
@joshkotrous Just gave the auth component a read--LGTM!

kylejryan (approved):

PR #146, opened 2026-02-17
fix: keep search bar visible in /sessions when no results match

### LINKED ISSUE #142: Allow typing in /sessions view when there are no results
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

When searching through sessions, if there are no results, the search bar disappears and you're forced to exit to the home screen and use the /sessions command again.

https://github.com/user-attachments/assets/71978c0a-a533-4c69-b856-fe4e92c3ddb6

We can continue to display the "no sessions found" text, but we should keep the search bar at the top so the user can keep typing and modify their search.

jorgeraad:
- When searching in `/sessions`, if all results are filtered out, the search bar now stays visible so users can modify their query
- The original empty state (no sessions at all) is preserved — no search bar shown when the user has zero sessions
- Adds `flexShrink={0}` to header, search, and footer to prevent layout shifts at smaller terminal sizes

Closes #142

### DIFF
--- a/src/tui/components/commands/sessions-browser.tsx
+++ b/src/tui/components/commands/sessions-browser.tsx
@@ -31,6 +31,7 @@ export default function SessionsBrowser() {
   const {
     groupedSessions,
     visualOrderSessions,
+    totalCount,
     loading,
     searchTerm,
     setSearchTerm,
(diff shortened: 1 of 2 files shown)


jorgeraad:
https://github.com/user-attachments/assets/9692ca05-7b65-479b-b596-bf0fd1dac13d

jorgeraad:
Original "No sessions" view is still shown when there are 0 sessions in the user's history:

https://github.com/user-attachments/assets/1a77b6d4-aebe-4842-804e-944a93f1381d

joshkotrous (approved):

PR #152, opened 2026-02-18
Fix discovery panel message duplication (#147)

### LINKED ISSUE #147: New messages in the pentest discovery panel are prefixed by all previous messages concatenated
### Description

When you run the /pentests command in the discovery panel, assistant messages are prefixed by all the previous messages concatenated, before getting the new input.

When the session is reloaded, this duplicated content is not persisted.

That said, there does seem to be some messages that are not persisted (e.g., "Initial Assessment...").

https://github.com/user-attachments/assets/04486bec-39c9-4ea8-ab17-fa829ea2b977

### Apex version

_No response_

### AI Provider

_No response_

### Steps to reproduce

_No response_

### Screenshots or logs

_No response_

### Operating System

_No response_

### Terminal

_No response_

jorgeraad:
Assistant messages in the discovery panel were prefixed with all previous messages concatenated. Two independent bugs caused this:

**1. Wrong stream event type (`"step-finish"` → `"finish-step"`)**
The stream handler checked for `"step-finish"` to reset the text accumulator between steps, but the AI SDK `fullStream` uses `"finish-step"` for step completion events ([docs](https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol), [ref](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text)). The check never matched, so `currentDiscoveryText` accumulated across the entire session.

Pentest agents weren't affected because the orchestrator manually constructs `SubAgentStreamEvent` objects with `type: "step-finish"` in its `onStepFinish` callback, bypassing the raw stream event name.

**2. Dual-write between `onDiscoveryStream` and `onDiscoveryStepFinish`**
Both handlers wrote assistant text to the same message array. The stream handler set content via accumulated deltas, then `onDiscoveryStepFinish` appended `step.text` (the same content) to the last assistant message.

With only fix 1, the old step's text briefly flashes in a new message bubble before being overwritten by the next step's first streaming token. Removing text-writing from `onDiscoveryStepFinish` eliminates that flash — tool call/result handling and token tracking remain.

Closes #147

### DIFF
--- a/src/tui/components/session-view/index.tsx
+++ b/src/tui/components/session-view/index.tsx
@@ -293,8 +293,9 @@ export default function SessionView({
                 step.usage.outputTokens ?? 0,
               );
 
-            // Update messages from step data (same pattern as onPentestAgentStream)
-            const { text, toolCalls, toolResults } = step;
+            // Update messages from step data — only tool calls and tool results.
+            // Text content is handled solely by onDiscoveryStream to avoid duplication.
+            const { toolCalls, toolResults } = step;
 
             setSubagents((prev) => {
               const idx = prev.findIndex(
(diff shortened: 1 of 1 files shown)


jorgeraad:
After fix:

https://github.com/user-attachments/assets/9cd5c9a7-69b6-49b4-96e8-292068de5463

joshkotrous (approved):

PR #160, opened 2026-02-19
fix: auto-scroll sessions browser on keyboard navigation (#156)

### LINKED ISSUE #156: Unable to scroll in sessions view
### Description

When you go down the list of sessions and are focused on a session that's out of the initial view, you should scroll automatically rather than remaining off screen.

https://github.com/user-attachments/assets/09ef1bd2-27a0-48c4-9d9c-f7c94ec1661f

### Apex version

_No response_

### AI Provider

_No response_

### Steps to reproduce

_No response_

### Screenshots or logs

_No response_

### Operating System

_No response_

### Terminal

_No response_

jorgeraad:
The `/sessions` browser wasn't scrolling to keep the selected item visible when navigating with arrow keys. Adapted the existing `scrollToIndex` logic from `sessions-display` so the scrollbox follows the selection — including wrapping from last→first and first→last.

Fixes #156

https://github.com/user-attachments/assets/14644b88-6a84-4f1e-b7d4-7a0e4239d3e4

### DIFF
--- a/src/tui/components/commands/help-dialog.tsx
+++ b/src/tui/components/commands/help-dialog.tsx
@@ -9,6 +9,7 @@
 import { useState, useEffect, useRef, useMemo } from "react";
 import { useKeyboard, useTerminalDimensions } from "@opentui/react";
 import { RGBA, ScrollBoxRenderable } from "@opentui/core";
+import { scrollToIndex } from "../../utils/scroll";
 import { useCommand } from "../../context/command";
 import { useRoute } from "../../context/route";
 import type { CommandConfig } from "../../command-registry";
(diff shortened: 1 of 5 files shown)


jorgeraad:
Also refactored the duplicated scroll-into-view logic that was duplicated across four components (sessions-browser, sessions-display, tools-panel, help-dialog).

I extracted it into src/tui/utils/scroll.ts. A utill file isn't the best long-term home for this though. I opened anomalyco/opentui#716 proposing a scrollChildIntoView method on ScrollBoxRenderable itself, which is where I think this logic belongs. If that doesn't get picked up, we could wrap it in a shared component instead and remove this util.

joshkotrous (approved):

PR #171, opened 2026-02-21
Implemented improved error handling

### LINKED ISSUE #166: Better error handling
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

We should add more robust and consistent error handling.

In some cases, apex crashes while in others, it fails silently and hangs as a result.

For example with the Bedrock issue #165, in operator mode, apex crashes and in pentest mode the discovery panel displays "thinking" indefinitely.

joshkotrous:
Implemented improved error handling within the TUI

TUI errors should no longer completely kill the TUI process, but are caught, displayed in an error toast, and written to `.pensar/error.log`

This way if users encounter an error, they have immediate feedback in the TUI with details, and we can ask them to share their error.log for further debugging. 

Fixes #166 

Where this should be improved upon is during agent execution, we should display errors properly in the TUI rather than silently hanging, and also write this to the error.log with an `[AGENT]` indicator so we know the error came form the agent.

### DIFF
--- a/src/core/logger/index.ts
+++ b/src/core/logger/index.ts
@@ -38,13 +38,15 @@ function pruneErrorLog(): void {
     const lines = raw.split("\n");
 
     const kept: string[] = [];
+    let keeping = true;
     for (const line of lines) {
       const match = TIMESTAMP_RE.exec(line);
       if (match) {
-        const ts = new Date(match[1]!).getTime();
-        if (ts < cutoff) continue;
+        keeping = new Date(match[1]!).getTime() >= cutoff;
       }
-      kept.push(line);
+      if (keeping) {
+        kept.push(line);
+      }
     }
 
     writeFileSync(ERROR_LOG_PATH, kept.join("\n"), "utf8");
--- a/src/core/logger/index.ts
+++ b/src/core/logger/index.ts
@@ -1,5 +1,12 @@
 import { sessions, type SessionInfo } from "../session";
-import { appendFileSync, existsSync, mkdirSync } from "fs";
+import {
+  appendFileSync,
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  writeFileSync,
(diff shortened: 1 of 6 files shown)


jorgeraad:
Confirmed it works as intended. Good improvement.

https://github.com/user-attachments/assets/c6df862c-305a-4ba1-9f88-ca7f007ae37e

Side note: realized a recent merge affected the theme. Will submit a fix.

jorgeraad (approved):
Good addition. ^Tested above.

Just had an agent do a pass on this PR and verify its results with tests. Could be worth taking a look:

1. Crash handlers don't log to the error log - The uncaughtException/unhandledRejection handlers call console.error but not writeErrorLog, so crashes won't show up in ~/.pensar/error.log. A writeErrorLog(err, "UNCAUGHT") call before process.exit would close that gap.
2. Log pruning can orphan stack traces - pruneErrorLog filters line-by-line, so when a timestamped entry is expired, its continuation lines (stack traces without timestamps) survive as orphans. A keeping flag that carries the decision forward to continuation lines would fix this.
3. Circuit breaker halts to a blank screen - When componentDidCatch hits the MAX_ERRORS threshold it sets halted: true and returns without notifying the user, so the UI just goes blank. Calling onError with a halt message before setting state would at least give the user a toast.

Nits:
- hasError branch in render is unreachable - getDerivedStateFromError sets hasError: true, but componentDidCatch immediately resets it to false (or sets halted), so the if (this.state.hasError) check in render can never actually be reached.
- Dual toast timers - The toast lifecycle has two independent timers (one in the context, one in the component) managing visibility. Works fine but could be simplified to a single source of truth.
- Unnecessary async on onMouseUp - The handler doesn't await anything so the async keyword isn't needed.

PR #187, opened 2026-02-25
Replace callback architecture with typed AgentEventBus and async-iterable AgentRun<T> API

jorgeraad:
Replaces the callback-based event plumbing (`ConsumeCallbacks`, `SubagentConsumeCallbacks`, dual dispatch in `consume()`) with two new primitives:

**AgentEventBus** — an internal typed pub/sub bus that agents `emit()` events onto. The key feature is `child()` buses: when a tool spawns a subagent, it creates `ctx.eventBus.child("agent-id")` and events automatically get tagged with the subagent ID and bubble up to the parent. This eliminates the copy-pasted callback forwarding wrappers that existed in `spawnPentestSwarm`, `runAttackSurface`, `spawnCodingAgent`, and `delegateAuth` — all of which manually constructed subagent callback objects with the same 15-line pattern.

**AgentRun\<T\>** — a public API wrapper that bridges the internal push-based bus to an external pull-based async iterable. Every `run*` function in `src/core/api/` now returns `AgentRun<T>` instead of `Promise<T>`. Clients can stream events with `for await (const event of run) { switch (event.type) { ... } }` or skip streaming entirely with `await run.result`. This is the same pattern used by the Vercel AI SDK's `fullStream` and Claude Code's `Query` — standard async iteration with a discriminated union.

What this deletes:
- `ConsumeCallbacks` and `SubagentConsumeCallbacks` type definitions (~60 lines)
- All callback forwarding wrappers in orchestration tools (~90 lines across 4 files)
- All callback unwrapping in the API layer
- Dual dispatch logic in `consume()` (which fired both the direct callback and the subagent callback for every event)
- `as never` type casts at forwarding boundaries

The `AgentEvent` discriminated union makes new event types additive — add a variant, emit it, handle it in the consumer's switch. No forwarding code to update.

For clients that currently `await` agent functions directly, the migration is appending `.result` to each call site. TypeScript catches every one since the return type changed.

A follow-up PR migrates the TUI components from `.on()` bus subscriptions to the same `for await` pattern the CLI uses, making the TUI just another API client.

### DIFF
--- a/scripts/attack-surface.ts
+++ b/scripts/attack-surface.ts
@@ -101,6 +101,16 @@ async function runAttackSurface(options: AttackSurfaceOptions): Promise<void> {
     });
 
     const eventBus = new AgentEventBus();
+    eventBus.on("text-delta", (e) => process.stdout.write(e.data.text));
+    eventBus.on("tool-call", (e) =>
+      console.log(
+        `\n→ calling ${e.data.toolName} \n ${JSON.stringify(e.data.input, null, 2)}`,
+      ),
+    );
+    eventBus.on("tool-result", (e) =>
+      console.log(`✓ ${e.data.toolName} completed`),
+    );
+    eventBus.on("error", (e) => console.error(e.error));
 
     const result = await runAttackSurfaceAgent({
       target: TARGET_URL,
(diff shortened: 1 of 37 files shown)


joshkotrous:
@jorgeraad for this one just pls ensure to open a pr in the console repo to ensure the integration works as expected / update it to adapt conventions

jorgeraad:
Superseded by #189 which consolidates all three event architecture PRs into a single review.

PR #189, opened 2026-02-25
Replace callback architecture with typed AgentEventBus and async-iterable AgentRun<T> API

jorgeraad:
## Summary

Replace Apex's callback-based event plumbing with a typed event bus internally and an async-iterable `AgentRun<T>` API externally. Every consumer (TUI, CLI, scripts, and future Console integration) now uses the same pattern:

```typescript
const run = runPentestAgent({ target, model, session });

for await (const event of run) {
  switch (event.type) {
    case "text-delta": ...
    case "tool-call": ...
    case "subagent-spawn": ...
  }
}

const { findings } = await run.result;
```

**Design doc:** `.jraad/docs/design-docs/20260224185615-unified-event-architecture.md`

---

### Layer 1: AgentEventBus (internal plumbing)

A typed pub/sub bus that replaces `ConsumeCallbacks` and `SubagentConsumeCallbacks`. The key innovation is hierarchical child buses:

```typescript
// Before — 15+ lines of boilerplate per orchestration tool
const result = await agent.consume({
  subagentCallbacks: ctx.subagentCallbacks ? {
    onTextDelta: (d) => ctx.subagentCallbacks!.onTextDelta?.({ ...d, subagentId }),
    onToolCall: (d) => ctx.subagentCallbacks!.onToolCall?.({ ...d, subagentId }),
    // ...
  } : undefined,
});

// After — 3 lines
const childBus = ctx.eventBus?.child(subagentId);
const agent = new TargetedPentestAgent({ ...input, eventBus: childBus });
const result = await agent.consume();
```

Events emitted on child buses are auto-tagged with `subagentId` and bubble up to the parent. No manual forwarding.

### Layer 2: AgentRun\<T\> (public API)

Wraps the bus in a standard `AsyncIterable<AgentEvent>`:
- **Stream events**: `for await (const event of run) { ... }`
- **Get result**: `await run.result`
- **Result only** (ignore streaming): `await runPentestAgent({ ... }).result`

All `run*` functions return `AgentRun<T>` instead of `Promise<T>`.

### Layer 3: TUI migration

Both TUI components (`pentest.tsx`, `operator-dashboard`) migrate from callback-based consumption to `for await` + `switch`. The TUI is now "just another client" with no privileged access to Apex internals.

### Layer 4: API export cleanup

- Expanded `src/core/api/index.ts` barrel to re-export all types Console needs (agent runners, event types, result types, input types, domain types)
- Console will import from two entry points: `src/core/api` and `src/core/session`
- No more deep imports into `src/core/agents/specialized/...`

---

## What gets deleted

- `ConsumeCallbacks` and `SubagentConsumeCallbacks` type definitions
- All callback forwarding wrappers in orchestration tools (~90 lines across 3 files)
- All callback unwrapping in the API layer
- All `as never` type casts at forwarding boundaries
- Dual dispatch logic in `consume()`

## What gets added

| Category | Files | Lines |
|----------|-------|-------|
| `AgentEventBus` + tests | 2 | +426 |
| `AgentRun` + tests | 2 | +463 |
| `runOperatorAgent` API | 1 | +32 |
| API barrel expansion | 1 | +40 |
| **Total new test lines** | | **~630** |

## Breaking change for Console

Every `run*Agent()` function now returns `AgentRun<T>` instead of `Promise<T>`. Console migration is mechanical:

```typescript
// Before
const { findings } = await runTargetedPentestAgent({ ... });
// After
const { findings } = await runTargetedPentestAgent({ ... }).result;
```

8 call sites across 7 files, all TypeScript-guided.

## Path to sandboxed agents

`AgentEvent` is a flat, serializable discriminated union. A server layer for remote sandboxes becomes trivially thin — wrap `for await` in an SSE endpoint. Consumer-side code is identical whether Apex runs in-process or remote.

### DIFF
--- a/scripts/attack-surface.ts
+++ b/scripts/attack-surface.ts
@@ -101,6 +101,16 @@ async function runAttackSurface(options: AttackSurfaceOptions): Promise<void> {
     });
 
     const eventBus = new AgentEventBus();
+    eventBus.on("text-delta", (e) => process.stdout.write(e.data.text));
+    eventBus.on("tool-call", (e) =>
+      console.log(
+        `\n→ calling ${e.data.toolName} \n ${JSON.stringify(e.data.input, null, 2)}`,
+      ),
+    );
+    eventBus.on("tool-result", (e) =>
+      console.log(`✓ ${e.data.toolName} completed`),
+    );
+    eventBus.on("error", (e) => console.error(e.error));
 
     const result = await runAttackSurfaceAgent({
       target: TARGET_URL,
--- a/scripts/attack-surface.ts
+++ b/scripts/attack-surface.ts
@@ -5,6 +5,7 @@ import { readFileSync, existsSync } from "fs";
 import { join } from "path";
 import { sessions } from "../src/core/session";
 import { runAttackSurfaceAgent } from "../src/core/api";
+import { AgentEventBus } from "../src/core/agents/offSecAgent";
 import { config } from "dotenv";
 
 config();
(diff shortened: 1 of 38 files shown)


jorgeraad:
Recreating with renamed branch refactor/event-driven-agent-api

PR #290, opened 2026-03-06
fix: searchbar input conflicts with navigation shortcuts

### LINKED ISSUE #157: Session mode conflict between searchbar input and key bindings
### Description

There are conflicts between the input text that goes into the search bar and the key bindings on the sessions page. For example:
- Typing the `o` in the search bar triggers the operator on whichever session is focused.
- The `r` never opens the focus session and instead is inserted into the search bar.
- The shortcuts key binding (`?`) isn't respected in the sessions view.

https://github.com/user-attachments/assets/9546ba1e-e211-4feb-b42d-db512613750c

### Apex version

_No response_

### AI Provider

_No response_

### Steps to reproduce

_No response_

### Screenshots or logs

_No response_

### Operating System

_No response_

### Terminal

_No response_

NucleiAv:
Fixes #157

## Problem

Pressing single-letter shortcuts (o, r) while typing in the sessions search input fires navigation actions instead of inserting characters.

- Pressing `o` while searching triggers "Open in Operator mode"
- Pressing `r` while searching triggers "Open Report"

## Fix

Track search input value in local state. When the input has text, single-letter shortcuts are suppressed. Arrow keys, Enter, and Escape continue to work normally while searching.

---

> [!NOTE]
> **Low Risk**
> Small UI keyboard-handling change scoped to the sessions dialog; low risk beyond potentially altering shortcut availability when a search term is present.
> 
> **Overview**
> Prevents single-letter navigation shortcuts in the sessions dialog from firing while the search input contains text.
> 
> `SessionsDisplay` now reads `searchTerm` from `useSessionsList` and gates the `O` (open in operator mode) and `R` (open report) keyboard handlers on `!searchTerm`, keeping arrow navigation, Enter, Escape, and Ctrl+D behavior unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 593afb3a797f21c97a488983719b0ab6162e05b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/tui/components/commands/sessions-display.tsx
+++ b/src/tui/components/commands/sessions-display.tsx
@@ -28,6 +28,7 @@ export default function SessionsDisplay({ onClose }: SessionsDisplayProps) {
     groupedSessions,
     visualOrderSessions,
     loading,
+    searchTerm,
     setSearchTerm,
     deleteSession: hookDeleteSession,
   } = useSessionsList();
@@ -99,11 +100,11 @@ export default function SessionsDisplay({ onClose }: SessionsDisplayProps) {
       return;
     }
 
-    // O - Open session in operator mode
     if (
       (key.name === "o" || key.name === "O") &&
       !key.ctrl &&
       !key.meta &&
+      !searchTerm &&
       visualOrderSessions.length > 0
     ) {
       const currentSelection = visualOrderSessions[selectedIndex];
@@ -143,8 +144,7 @@ export default function SessionsDisplay({ onClose }: SessionsDisplayProps) {
       return;
     }
 
-    // R - Open report
-    if (key.name === "r" && visualOrderSessions.length > 0) {
+    if (key.name === "r" && !searchTerm && visualOrderSessions.length > 0) {
       const currentSelection = visualOrderSessions[selectedIndex];
       if (!currentSelection) return;
       openReport(currentSelection.id);


jorgeraad:
@NucleiAv, tested and your proposed change and confirmed it works.

That said, I'm not sure we are adding a new `useRef` here, we already have the `searchTerm` state. Please remove, this change should just modify the condition on the if statement to not propagate the key if there is search text.

Nit: no need for a comment explaining. This is a simple change and the code speaks for itself.

___

Once that is updated, I will merge since it is clearly an improvement over the current behavior. That said, I will keep the issue open as this we still need to improve the UX here.

The likely long-term solution is to either:

1. change our shortcuts to use modifier keys (e.g., ctrl), in which case we should apply this to other shortcuts in Apex for consistency OR
2. Introduce a visual "selected" state to this view, where either the searchbar or a session item is clearly in focus. When the searchbar is in focus, it should be clear that any keys are simply inputs to the text input. If the user scrolls down and has a session item focused, pressing "o" or "r" will perform their mapped functions instead of inputting into the searchbar. A nice UX detail here would be: if the pressed key doesn't have a binding, then assume they want to type in the search bar and auto-select and pass it into the searchbar (thus unselecting the session item).

NucleiAv:
@jorgeraad Updated. Removed the `searchValueRef`, now reading `searchTerm` directly from `useSessionsList` (it was already in the hook's return value). Simplified `onInput` back to `setSearchTerm` and dropped the explanatory comments.

NucleiAv:
@jorgeraad i have already made changes 3 days ago... kindly review.
lemme know if any thing is needed from my end!

PR #291, opened 2026-03-06
fix: /sessions resume starts fresh instead of continuing

### LINKED ISSUE #289: /sessions command doesn't work
### Description

When running /sessions and trying to resume a previous session, it just restarts the session for some reason but does not actually pull it up. 

### Apex version

_No response_

### AI Provider

_No response_

### Steps to reproduce

_No response_

### Screenshots or logs

_No response_

### Operating System

_No response_

### Terminal

_No response_

NucleiAv:
Fixes #289

## Problem

Resuming an interrupted session restarts the pentest agent from scratch instead of showing where the session left off.

## Fix

The changes:
1. Interrupted sessions now set phase to "pentesting" to show saved state instead of calling startPentest()
2. Added isExecuting prop to OrchestratorPanel
3. isRunning in OrchestratorPanel now requires isExecuting to be true, so the spinner only shows when the agent is actually running
4. Infer resumed phase from saved subagents: sessions interrupted during discovery now correctly restore to "discovery" phase rather than "pentesting", fixing a misleading phase label
5. Infer resumed phase from saved subagents: sessions interrupted during discovery correctly restore to "discovery" phase
6. Gate discovery and reporting spinners on isExecuting to prevent infinite spinner animation on resumed sessions where no agent runs

## Behaviour after fix

- Completed sessions: read-only view (unchanged)
- Interrupted sessions: shows saved agent cards and discovery logs without restarting
- New sessions: start immediately (unchanged)

---

> [!NOTE]
> **Medium Risk**
> Touches pentest session restore/phase state machine and execution indicators; mistakes could mislabel phases or hide/show spinners incorrectly when resuming sessions.
> 
> **Overview**
> Fixes session resume so loading an interrupted run from `/sessions` *restores saved subagent state* instead of calling `startPentest()` and starting over.
> 
> `Pentest` now infers the initial UI `phase` from persisted subagent status (discovery vs pentesting vs reporting), treats `isComplete` without `hasReport` as reporting-incomplete, and gates orchestrator/discovery/reporting spinners via a new `isExecuting` prop so resumed-but-paused sessions don’t show infinite “running” indicators. Separately, session opening no longer infers operator mode from `hasOperatorState`, only from `config.mode`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fc15fd2efedf076d7b04e858e5987778c5684892. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/tui/components/commands/sessions-display.tsx
+++ b/src/tui/components/commands/sessions-display.tsx
@@ -86,9 +86,7 @@ export default function SessionsDisplay({ onClose }: SessionsDisplayProps) {
         setTimeout(() => setStatusMessage(""), 2000);
         return;
       }
-      const isOperator =
-        currentSelection.config?.mode === "operator" ||
-        currentSelection.hasOperatorState;
+      const isOperator = currentSelection.config?.mode === "operator";
       refocusPrompt();
       onClose();
       route.navigate({
--- a/src/tui/components/pentest/pentest.tsx
+++ b/src/tui/components/pentest/pentest.tsx
@@ -211,22 +211,49 @@ export default function Pentest({
             setPentestAgents(agents);
           }
 
-          if (state.hasReport || state.isComplete) {
-            // Truly finished — show read-only completed view
+          if (state.hasReport) {
+            // Truly finished with report — show read-only completed view
             setPhase("completed");
             if (attackSurfaceAgents.some((sa) => sa.messages.length > 0)) {
               setShowOrchestratorPanel(true);
             }
-          } else {
-            // Session was interrupted — show saved state then restart
+          } else if (state.isComplete) {
+            // All pentest agents done but report was not generated
+            // (interrupted during reporting) — show reporting phase
             if (attackSurfaceAgents.some((sa) => sa.messages.length > 0)) {
               setShowOrchestratorPanel(true);
             }
-            startPentest(s);
+            setPhase("reporting");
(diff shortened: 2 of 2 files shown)


jorgeraad:
Hey @NucleiAv! Thanks for opening the PR. I tested your change and it doesn't seem to fix the issue. Would you be able to include a screen recording on this and any future fixes?

I'm looking into the issue as well.

NucleiAv:
@jorgeraad seems to be working fine at my end.

https://github.com/user-attachments/assets/36d37eeb-4f8a-4541-a2d6-31d40a9c06b5

jorgeraad:
Hi @NucleiAv, the behavior in your screen recording is actually the same as the current behavior I already see in the canary branch. So, there are two problems with this that we need to fix:
1. By default, if you open an unfinished pentest session, it should reopen the pentest mode. Right now, everything is being opened in operator mode, which should not be the default (but you should still be able to do by pressing "o").
2. When it is changed so that it opens in pentest mode, it should resume where it left off. It should not start over and it should not appear as completed with an empty pentest report.

NucleiAv:
@jorgeraad i have already made changes 2 days ago... kindly review.
lemme know if any thing is needed from my end!

PR #301, opened 2026-03-09
Lazy-discover Lambda Function URL for SSE streaming

Yuvanesh-ux:
## Summary
- Adds `streamUrl` config option to `PensarModelConfig` for explicit override
- On first `doStream()` call, queries `GET /bedrock/validate` for the Lambda Function URL (`endpoints.stream`) and caches it
- Falls back to buffered `POST /bedrock/stream` API Gateway route if Lambda URL unavailable

## Test plan
- [ ] Set `PENSAR_DEBUG=1`, run Apex — confirm "Discovered stream URL: https://...lambda-url..." in logs
- [ ] Verify SSE streaming works token-by-token via Lambda Function URL
- [ ] If validate returns no stream URL, confirm fallback to `/bedrock/stream`

---

> [!NOTE]
> **Medium Risk**
> Changes request routing and payloads for streaming, which could impact runtime behavior if the validate response/stream URL is misconfigured or differs across environments.
> 
> **Overview**
> Updates the Pensar provider’s `doStream()` to **lazily discover and cache** the true SSE streaming endpoint by calling `/bedrock/validate` and using `endpoints.stream` when available.
> 
> If discovery fails, streaming requests fall back to posting to `/bedrock/invoke` with `stream: true` (and still fall back to wrapping `doGenerate()` on fetch/errors), changing the default streaming route away from the previous hardcoded `/bedrock/stream`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 69b71f5b30a53609d58aad2f93d15401e2f568e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/core/ai/providers/pensar.ts
+++ b/src/core/ai/providers/pensar.ts
@@ -35,6 +35,8 @@ export interface PensarModelConfig {
    * allowing transparent token refresh for WorkOS auth.
    */
   getToken?: () => Promise<{ token: string; type: "workos" | "legacy" } | null>;
+  /** Optional explicit stream URL. If not set, discovered via /bedrock/validate. */
+  streamUrl?: string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


joshkotrous:
@Yuvanesh-ux  we should only have 1 url so this doesnt get confusing, right now it looks like theres both `baseUrl` and `streamUrl`. The only one we should support is the streaming one. 

I also think in console we have a redundant function because when we were troubleshooting the streaming functionality an additional lambda got added, perhaps im wrong but something to double check.

Yuvanesh-ux:
@joshkotrous 

In dev (npx sst dev):                                                                                                     
  - POST /bedrock/invoke with stream: true → buffered SSE
  - API Gateway collects all Bedrock chunks, formats them as SSE events, returns the entire body at once                    
  - Text appears all at once after Bedrock finishes — no token-by-token effect                          
  - This is the only option because Lambda Function URLs with awslambda.streamifyResponse don't work in SST local dev       
                                                                                                                            
  In production (deployed):                                                                                                 
  - Lambda Function URL → true SSE streaming                                                                                
  - Bedrock chunks are written to the response stream as they arrive — real token-by-token                                  
  - The Apex CLI discovers this URL from GET /bedrock/validate (the endpoints.stream field)
  - The validate handler only returns this URL when NODE_ENV !== 'development'                                              
                                                                                                                            
  How Apex picks the right one:                                                                                             
  1. Calls /bedrock/validate → checks for endpoints.stream in the response                                                  
  2. If present (production) → uses the Lambda Function URL                                                                 
  3. If absent (dev) → falls back to /bedrock/invoke with stream: true                                                      
                                                                                                                            
  Both return the same SSE protocol (event: type\ndata: json\n\n + pensar:usage at the end), so the Apex parser works       
  identically for both. The only difference is latency — buffered vs real-time delivery

joshkotrous:
@Yuvanesh-ux I get why this works, but I don’t think it’s the right abstraction. We shouldn’t expose both baseUrl and streamUrl in the model config. Apex should have one canonical endpoint contract, and dev/prod differences should be handled internally. Otherwise we’re baking temporary backend complexity into the client interface.

Yuvanesh-ux:
@joshkotrous addressed your issue, still need the lambda though so we get true streaming when deployed to prod. But all routing and such is handled in backend now instead of client exposed

joshkotrous (approved):

PR #406, opened 2026-03-17
CLI: support node (with helpful bun dependency error for tui) and unify npm and binary entry points into single router

jorgeraad:
`bin/pensar.js` and `src/cli.ts` were two independent copies of the same CLI router — duplicated help text, duplicated command routing, duplicated upgrade logic. They'd already drifted out of sync: bin had dead commands (`benchmark`, `swarm`, `quicktest`) routing to build files that didn't exist, while cli.ts had `targeted-pentest` and `doctor` that bin didn't know about.

This eliminates the duplication by building `src/cli.ts` into `build/cli.js` and reducing `bin/pensar.js` to a thin wrapper (282 → 35 lines). `src/cli.ts` is now the single source of truth for all CLI routing and help text. Adding a new command means adding it in one place.

### Node + Bun runtime support

The npm package now works under both Node (>=18) and Bun (>=1.0). Previously the shebang was `#!/usr/bin/env bun` but the `engines` field claimed Node support — in reality, the package crashed immediately under Node due to Bun-specific syntax (JSON imports without `with { type: "json" }`, raw `.ts` imports).

The shebang is now `#!/usr/bin/env node` so it runs everywhere. Runtime behavior:

- **CLI commands** (`pensar --help`, `pensar auth`, `pensar issues`, etc.) — work natively under both Node and Bun. No Bun required.
- **TUI** (bare `pensar` with no args) — requires Bun because of `@opentui`'s `bun:` protocol dependencies. When running under Node, `bin/pensar.js` detects the runtime and automatically re-execs itself under Bun if available. If Bun isn't installed, it shows a clear error: *"TUI mode requires Bun. Install Bun (https://bun.sh) or use a standalone binary release."*
- **Compiled binary** — unaffected by all of this. It bundles everything including the Bun runtime.

`Bun.spawn` in `auth.ts` (used to open the browser during `pensar auth login`) was replaced with `child_process.spawn` so it works in both runtimes.

### Build changes

- Build uses `--splitting` so TUI-related chunks are code-split and never loaded under Node
- Externalized `react`, `react/jsx-runtime`, `react/jsx-dev-runtime`, and `react-reconciler` — fixes a duplicate React instance crash where the build bundled its own React while `@opentui/react` used `node_modules/react`
- Removed the separate `bun build src/tui/index.tsx` step — now handled via cli.ts's code splitting
- Removed `src/core/installation` from npm `files` — bundled into `build/cli.js` now (was also shipping the test file unnecessarily)
- Updated `main` field from `build/index.js` to `build/cli.js`
- Consolidated all `console.log`-per-line help text into template literals
- Build output: 55 files / 22MB → 43 files / 8.2MB

### CI

Smoke tests for all three distribution paths: npm package under Node (including TUI error message assertion with Bun stripped from PATH), npm package under Bun, and compiled binary.

### Test plan

tested manually and confirmed:
```
⏺ cd ~/Projects/apex-worktrees/apex-cli-router-unify
  rm -rf build/ && bun run build
  npm pack && npm install -g ./pensar-apex-0.0.104.tgz && rm pensar-apex-0.0.104.tgz

  # Node — CLI commands
  node $(which pensar) --version
  node $(which pensar) --help
  node $(which pensar) auth --help
  node $(which pensar) pentests --help

  # Node — TUI (should re-exec under Bun automatically)
  node $(which pensar)

  # Node — TUI without Bun (should show error)
  PATH=$(echo "$PATH" | tr ':' '\n' | grep -v bun | tr '\n' ':') node $(which pensar)

  # Bun — everything
  pensar --version
  pensar --help
  pensar auth --help
  pensar
```

Also added these checks to the github actions build step.

---

> [!NOTE]
> **Medium Risk**
> Changes the published package entrypoint, build output, and runtime behavior (Node vs Bun) for all CLI invocations, so a regression could break installs or TUI launch behavior across environments. Logic changes are mostly routing/packaging and CI smoke coverage reduces risk.
> 
> **Overview**
> **Unifies the CLI entrypoint and adds first-class Node support for the published npm package.** `bin/pensar.js` is reduced to a thin Node wrapper that forwards to `build/cli.js`, re-execing under Bun only for TUI mode (no args) and otherwise setting `PENSAR_NO_TUI` to prevent accidental TUI loads.
> 
> **Build/packaging is reoriented around `src/cli.ts` as the single router source of truth.** `package.json` now points `main` to `build/cli.js`, the `build` script compiles `src/cli.ts` with code-splitting and externals for OpenTUI/React, and npm published files are trimmed accordingly.
> 
> **CI now smoke-tests all distribution paths.** The workflow installs the packed npm tarball and validates CLI help/version under Node and Bun, asserts a helpful error when Bun is unavailable for TUI mode, and also exercises the compiled binary path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4f1ec5abcbd181324f9153e3cf75bde22d18a4d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,8 +28,43 @@ jobs:
       - name: Build
         run: bun run build
 
-      - name: Smoke test published package
+      # Simulate `npm install -g @pensar/apex` — all smoke tests below use this
+      - name: Install npm package globally
         run: |
           npm pack
           npm install -g ./pensar-apex-*.tgz
-          pensar --version
+
+      # CLI commands should work natively under Node without Bun
+      - name: "Smoke: npm package under Node"
+        run: |
+          set -euo pipefail
+          node $(which pensar) --version
+          node $(which pensar) --help
+          node $(which pensar) auth --help
+          node $(which pensar) pentests --help
+
+      # Bare `pensar` under Node with no Bun on PATH should show a helpful error
+      - name: "Smoke: npm TUI fallback when Bun unavailable"
+        run: |
+          set -euo pipefail
+          PATH=$(echo "$PATH" | tr ':' '\n' | grep -v bun | tr '\n' ':') \
+            node $(which pensar) 2>/tmp/pensar-tui.txt || true
+          grep -q "TUI mode requires Bun" /tmp/pensar-tui.txt
+
+      # Explicitly invoke under Bun — bypasses the #!/usr/bin/env node shebang
+      - name: "Smoke: npm package under Bun"
+        run: |
+          set -euo pipefail
+          bun $(which pensar) --version
+          bun $(which pensar) --help
+          bun $(which pensar) auth --help
+          bun $(which pensar) pentests --help
+
+      # Standalone binary — self-contained, no Node/Bun runtime needed
+      - name: "Smoke: compiled binary"
+        run: |
+          set -euo pipefail
+          bun run build:binary
+          ./pensar --version
+          ./pensar --help
+          ./pensar auth --help
--- a/README.md
+++ b/README.md
@@ -20,11 +20,9 @@ Want to run from the cloud or integrate it with your CI/CD? See <a href="https:/
   <img src="screenshot.png" alt="Pensar Apex Screenshot" width="800">
 </p> -->
 
-
 ## Use Cases
 
 Apex enables both developers and security professionals to run autonomous and assisted penetration testing directly from the terminal.
-
 
 ### Developers: Run a Pentest in Minutes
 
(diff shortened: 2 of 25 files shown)


joshkotrous (approved):
Good w/ this as long as it doesnt break the build and core `pensar` command

I think thats where the .js file came from, was to enable the pensar command at one point after build

jorgeraad:
Fixes #425

PR #486, opened 2026-03-20
Feat/threat model

jorgeraad:
implement threat model agent


jorgeraad:
draft of an alternative approach.

PR #534, opened 2026-04-02
Adaptive Thinking Toggle

kylejryan:
## Summary

Adds extended thinking support to the pentest agent pipeline, enabling Apex to reason through complex security decisions before acting. This uses **adaptive thinking mode**, which lets the model decide per-turn how much reasoning to apply based on the complexity of the task at hand.

### Which agents use thinking?

Extended thinking is available to all agent types, but in practice **pentest agents are the primary beneficiaries**. The model naturally adapts its reasoning to the task complexity:

| Agent type | Reasoning usage | Why |
|-----------|----------------|-----|
| **Pentest agents** | 33-86% of steps | Complex security decisions — choosing attack vectors, analyzing responses for vulnerabilities, crafting exploit payloads |
| **Orchestrator** | Initial planning step | High-level session planning and delegation |
| **Attack surface agents** | 0% | Straightforward enumeration (crawl, probe, catalog) — the model correctly determines no deep reasoning is needed |

This distribution is exactly what we want. Attack surface mapping is mechanical — thinking tokens would be wasted. Pentest agents face genuinely ambiguous decisions where reasoning produces better vulnerability discovery.

### Why adaptive thinking? (and what we tried first)

Penetration testing agents face a mix of trivial steps (reading a tool result, issuing the next HTTP request) and genuinely complex decisions (choosing an attack vector, analyzing a response for subtle vulnerabilities, deciding whether a behavior is exploitable). We went through several iterations to find the right approach.

**Attempt 1: `type: "enabled"` with a fixed 10k token budget**

Our first implementation used a fixed thinking budget of 10,000 tokens per turn. In W&B traces, the model only produced reasoning on step 0 of each agent (the initial planning step) and then went silent — 1 out of 305 records had any reasoning content. The fixed budget was too small to be worth the model's effort on subsequent tool-use steps, so it effectively learned to skip thinking entirely after the first turn. We also found that `max_tokens` must exceed `budget_tokens` (Anthropic rejects the request otherwise), and temperature must be omitted — constraints that added complexity to the provider layer for marginal benefit.

**Attempt 2: `type: "adaptive"`**

Switching to adaptive mode removed the budget constraint entirely and let the model decide per-turn whether and how much to think. Results improved dramatically:

| Metric | Fixed budget (10k) | Adaptive |
|--------|-------------------|----------|
| Steps with reasoning | 11/382 (2.9%) | 207/368 (56.2%) |
| Total reasoning chars | 1,405 | 173,082 |
| Agents reasoning beyond step 0 | 0/12 | 12/12 |
| Per-agent reasoning rate | 0-4% | 33-86% |

The model now reasons deeply when the problem warrants it (analyzing whether a 403 response is bypassable, deciding which injection payload to try next) and skips thinking on mechanical steps (reading command output, issuing the next request in a sequence). This is exactly the behavior we want — reasoning where it matters, no wasted tokens where it doesn't.

We kept the `type: "enabled"` code path in the Pensar provider formatter for completeness (callers other than `streamResponse` could use it), but the default and recommended mode is adaptive.

### Verified on both providers

We validated thinking end-to-end on both the direct Anthropic API and the Pensar hosted inference gateway:

| Provider | Steps with reasoning | Reasoning chars | Per-agent range |
|----------|---------------------|-----------------|-----------------|
| **Direct Anthropic** | 207/368 (56.2%) | 173,082 | 33-86% |
| **Pensar Gateway** | 120/274 (43.8%) | 59,816 | 38-67% |

Both show the same pattern: pentest agents reason heavily, attack surface agents skip reasoning entirely, and the orchestrator reasons during initial planning. The Pensar gateway passes thinking params and response blocks through cleanly.

### How to configure

1. **Toggle in the UI**: Open the model picker (either from the config screen or via the `/model` command). An "Extended Thinking" checkbox appears below the model list. Toggle it on for any supported model.

2. **Persisted to config**: The setting is saved to `~/.pensar/config.json` as `reasoningEnabled: true` and persists across sessions.

3. **Supported models**: The selected model must support adaptive thinking. The checkbox is disabled (with a "not supported" label) for models that don't.

4. **Works with all providers**: Direct Anthropic, AWS Bedrock, and Pensar Gateway all support extended thinking. The Pensar provider now handles the full thinking round-trip including `thinking`/`thinking_delta`/`signature_delta` SSE events and multi-turn signature forwarding.

### Extended thinking pipeline
- **`src/core/ai/ai.ts`** — `modelSupportsThinking()` function, `providerOptions` construction with adaptive thinking mode, `enableThinking` option on `StreamResponseOpts`
- **`src/core/ai/providers/pensarFormatters.ts`** — Forwards `thinking` config to Bedrock request body (both `enabled` and `adaptive` modes), handles `reasoning` content parts in assistant messages for multi-turn round-trip (including signature), parses `thinking` response blocks into `reasoning` content
- **`src/core/ai/providers/pensar.ts`** — SSE stream handler emits `reasoning-start`/`reasoning-delta`/`reasoning-end` for thinking blocks, captures `signature_delta` for multi-turn, `doGenerate` collects reasoning parts

### Flag threading (all agents)
- **`OffensiveSecurityAgentInput`**, **`SpecializedAgentInput`**, **`ToolContext`** — `enableThinking?: boolean`
- **All 8 specialized agents** — destructure and forward `enableThinking` to `super()`
- **Orchestration tools** (`runPentestWorkflow`, `spawnPentestSwarm`, `spawnCodingAgent`) — forward from `ToolContext`
- **Workflow layer** (`pentest.ts`) — thread through `PentestWorkflowInput` and `PentestSwarmInput`

### UI
- **`ModelPicker.tsx`** — "Extended Thinking" checkbox with model support detection
- **`AgentProvider`** (`context/agent.tsx`) — `reasoningEnabled` state persisted to config
- **`ConfigView`**, **`ModelPickerDialog`** — pass reasoning props through

### Cleanup (separate from thinking)
- Removed plan mode tools (`writePlan`, `submitPlan`) and all associated UI/state
- Fixed W&B test that failed when `WANDB_PROJECT` env var was set

## Test plan

- [x] Verified with W&B Weave traces: 56% of steps produce reasoning with adaptive mode (direct Anthropic)
- [x] Verified with W&B Weave traces: 44% of steps produce reasoning via Pensar Gateway hosted inference
- [x] Pentest agents reason on 33-86% of steps; attack surface agents correctly skip reasoning (0%)
- [x] All pentest sub-agents receive and use thinking (confirmed via per-agent trace breakdown)
- [x] Pensar Gateway passes thinking params and response blocks through cleanly (confirmed end-to-end)
- [x] `bunx tsc --noEmit` — zero type errors
- [x] `bun run lint` — clean
- [x] `bun run format:check` — clean
- [x] W&B upload tests pass
- [x] Verify thinking toggle persists correctly across app restarts

Screenshot 2026-04-02 at 3 55 45 PM

Screenshot 2026-04-02 at 3 56 35 PM

Screenshot 2026-04-02 at 3 56 12 PM

---

> [!NOTE]
> **Medium Risk**
> Touches the AI provider layer and threads a new `enableThinking` flag through multiple agent/workflow entry points, which could affect inference request formatting and streaming parsing across providers/models.
> 
> **Overview**
> Adds an **“Extended Thinking” (adaptive reasoning)** capability that can be toggled in the TUI and persisted to config (`reasoningEnabled`). The toggle is shown only for supported Claude models via new `modelSupportsThinking()` detection.
> 
> Threads `enableThinking` through the offensive-security agent harness, orchestration tools, workflows, and specialized agents so sub-agents inherit the setting; the operator dashboard passes the UI flag into agent creation.
> 
> Updates the AI layer to pass Anthropic `providerOptions` for adaptive thinking when enabled, and extends the Pensar provider/formatters to round-trip `thinking` blocks (including signature) by emitting/parsing `reasoning-*` stream events and mapping them into request/response message formats. Includes a small test fix to ensure default W&B project behavior when `WANDB_PROJECT` is set.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a7ccdf85aeafb4f7b58e4f873696fe369985908c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/core/agents/offSecAgent/offensiveSecurityAgent.ts
+++ b/src/core/agents/offSecAgent/offensiveSecurityAgent.ts
@@ -170,6 +170,7 @@ export class OffensiveSecurityAgent<TResult = void> {
       persistentShell: this.persistentShell,
       skillsRegistry: input.skillsRegistry,
       traceWriter,
+      enableThinking: input.enableThinking,
     });
 
     let tools: ToolSet = input.extraTools
@@ -354,6 +355,7 @@ export class OffensiveSecurityAgent<TResult = void> {
         };
         input.onCacheMetrics?.(metrics);
       },
+      enableThinking: input.enableThinking,
       silent: true,
     });
   }
--- a/src/core/agents/offSecAgent/tools/runPentestWorkflow.ts
+++ b/src/core/agents/offSecAgent/tools/runPentestWorkflow.ts
@@ -60,6 +60,7 @@ Omit \`cwd\` for blackbox mode (live target probing only).`,
           authConfig: ctx.authConfig,
           abortSignal: ctx.abortSignal,
           eventBus: ctx.eventBus,
+          enableThinking: ctx.enableThinking,
         });
 
         const elapsedSeconds = Math.round((Date.now() - startedAt) / 1000);
--- a/src/core/agents/offSecAgent/tools/spawnCodingAgent.ts
+++ b/src/core/agents/offSecAgent/tools/spawnCodingAgent.ts
@@ -206,6 +206,7 @@ async function runSingleCodingAgent(
     abortSignal: ctx.abortSignal,
     eventBus: localBus,
     subagentId,
+    enableThinking: ctx.enableThinking,
   });
 
   try {
--- a/src/core/agents/offSecAgent/tools/spawnPentestSwarm.ts
+++ b/src/core/agents/offSecAgent/tools/spawnPentestSwarm.ts
@@ -90,6 +90,7 @@ Pass every target you want tested — the swarm handles concurrency automaticall
         findingsRegistry,
         eventBus: ctx.eventBus,
         concurrency: DEFAULT_CONCURRENCY,
+        enableThinking: ctx.enableThinking,
       });
 
       // Aggregate
--- a/src/core/agents/offSecAgent/tools/types.ts
+++ b/src/core/agents/offSecAgent/tools/types.ts
@@ -82,4 +82,7 @@ export type ToolContext = {
    * When present, checkpoint_state tool is available.
    */
   traceWriter?: StepTraceWriter;
+
+  /** Enable extended thinking for sub-agents spawned by orchestration tools. */
+  enableThinking?: boolean;
 };
--- a/src/core/agents/offSecAgent/types.ts
+++ b/src/core/agents/offSecAgent/types.ts
@@ -239,6 +239,9 @@ export type OffensiveSecurityAgentInput<TResult = void> = {
    * never see them.
    */
   environmentVariables?: Record<string, string>;
+
+  /** Enable extended thinking (reasoning) for supported models. */
+  enableThinking?: boolean;
 };
 
 /**
@@ -303,6 +306,9 @@ export interface SpecializedAgentInput {
    * Forwarded to the underlying {@link OffensiveSecurityAgentInput}.
    */
   environmentVariables?: Record<string, string>;
+
+  /** Enable extended thinking (reasoning) for supported models. */
+  enableThinking?: boolean;
 }
 
 /**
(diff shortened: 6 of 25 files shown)


Yuvanesh-ux:
@kylejryan This is anthropic only correct? If so, have you seen any regression when using non anthropic models?

kylejryan (commented):
BAD CURSOR! You're breaking the spec

Yuvanesh-ux (approved):

PR #540, opened 2026-04-03
Judge should flag title inflation where claimed impact exceeds demonstrated impact

Yuvanesh-ux:
### What does this PR do?

This PR implements title accuracy validation in the finding judge to prevent pentest agents from writing finding titles that claim full exploit chains while the POC only demonstrates step 1.

**Problem:**
Pentest agents write inflated titles like:
- *"Missing Rate Limiting on OTP Verification Endpoint **Enables Password Reset Takeover**"* — POC only demonstrates 100 requests without 429 responses. No account was actually taken over.
- *"Missing Rate Limiting on Invitation Code Acceptance **Enables Brute-Force**"* — POC sends 50 requests without rate limiting. No code was actually brute-forced.

These titles assert "enables [full attack]" but the POC only proves "missing [defense]". A rate limit test ≠ account takeover.

**Root Cause:**
The judge's "Severity Alignment" criterion checked for gross mismatches ("a reflected XSS POC should not claim RCE") but didn't check whether the title's claimed impact matches what was actually *demonstrated* vs. what *could theoretically happen*.

**Solution:**

1. **Enhanced Severity Alignment criterion** with explicit title inflation red flags:
   - Title claims multi-step attack but POC only demonstrates enabling condition
   - Title says "Enables X" but POC doesn't demonstrate X succeeding
   - Specific examples: rate limiting tests shouldn't claim takeover, user enumeration shouldn't claim auth bypass

2. **Added title validation fields** to judge output:
   ```typescript
   titleAccurate: boolean  // Whether title reflects demonstrated impact
   suggestedTitle?: string // Corrected title when titleAccurate is false
   ```

3. **Auto-apply suggested titles** in `documentFinding.ts`:
   - When judge flags title inflation, the suggested title replaces the original
   - Title correction is recorded in finding metadata
   - Markdown reports include a note when titles were corrected

**Changes:**
- `src/core/agents/specialized/findingJudge/index.ts`:
  - Enhanced criterion #4 with title accuracy validation rules
  - Added `titleAccurate` and `suggestedTitle` to schema and result type
  - Updated system prompt with title validation guidelines
- `src/core/agents/offSecAgent/tools/documentFinding.ts`:
  - Apply suggested title when judge flags inflation
  - Store title correction metadata in finding JSON
  - Add visual indicator in markdown when title was corrected
- `src/core/agents/specialized/findingJudge/index.test.ts`:
  - New integration tests demonstrating title validation

**Automated Testing:**
1. All existing tests pass (583 tests across 30 test files)
2. TypeScript compilation succeeds with no errors
3. ESLint passes with no violations
4. Added comprehensive integration tests:
   - Test 1: Detects "Password Reset Takeover" inflation → suggests "Missing Rate Limiting on OTP Verification Endpoint"
   - Test 2: Accepts accurate "SQL Injection" title when POC demonstrates actual exploit
   - Test 3: Validates title accuracy evaluation for borderline cases

**Test Results:**
```
✓ Finding Judge - Title Accuracy > should flag title inflation when title claims full attack but POC only demonstrates enabling condition [1813ms]
  - titleAccurate: false
  - suggestedTitle: "Missing Rate Limiting on OTP Verification Endpoint"
  
✓ Finding Judge - Title Accuracy > should accept accurate title when POC demonstrates actual exploit [1778ms]
  - titleAccurate: true
  - Valid: true
```

**Example Correction:**
Original inflated title:
```
"Missing Rate Limiting on OTP Verification Endpoint Enables Password Reset Takeover"
```

Judge's corrected title:
```
"Missing Rate Limiting on OTP Verification Endpoint"
```

Judge's reasoning:
> "The POC only demonstrates the lack of rate limiting, not that an attacker could actually take over accounts through password reset. The title and description overclaim the impact."

The implementation ensures titles describe what was **DEMONSTRATED**, not theoretical downstream impact. Theoretical impact belongs in the impact/description fields where it can be properly contextualized.

<div><a href="https://cursor.com/agents/bc-3510a887-725c-42bb-b1f3-800efb919106"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3510a887-725c-42bb-b1f3-800efb919106"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


JiwaniZakir (commented):
The mutation of `input.title` in `documentFinding.ts` (around the `if (!judgeResult.titleAccurate && judgeResult.suggestedTitle)` block) silently modifies the caller's object in place, which is a side effect that could surprise callers who inspect `input` after the call. It would be safer to use the corrected title explicitly downstream rather than mutating the input. Relatedly, the markdown is built using `# ${finding.title}` — if `finding` and `input` are separate objects (worth confirming), the corrected title won't appear in the markdown header even though the "Note" metadata line says the title was corrected, creating a contradiction in the output artifact. In `index.test.ts`, test 2's assertion `expect(result.suggestedTitle).toBeUndefined()` is fragile against an LLM returning `null` or an empty string instead of omitting the field; asserting falsiness (`expect(result.suggestedTitle).toBeFalsy()`) would be more robust. Test 3's claim title — "Missing Rate Limiting on Invitation Code Acceptance Enables Brute-Force" — may not actually be flagged by the judge since it accurately describes the enabling condition rather than claiming the brute-force was demonstrated, making this a potentially incorrect test expectation that could produce a false pass depending on the model's interpretation.

Yuvanesh-ux:
@JiwaniZakir 

Thanks for the review! A few thoughts:                                                                                 
                                                                                                                         
Input mutation — Fair point, though in practice the input object isn't reused after this tool executes. Happy to       
refactor if you feel strongly.                                                                                         
                                                                                                                       
Markdown header contradiction — This one isn't actually an issue. The mutation happens at line 217, before the finding 
object is created at line 334 with title: input.title. So by the time we hit # ${finding.title}, it already has the    
corrected title. No contradiction.                                                                                     
                                                                                                                       
toBeUndefined() fragility — The judge output is validated through a Zod schema, so the field is either present or      
undefined — never null or empty string.                                                                                
                                                                                                                       
Test 3 false pass — The test intentionally doesn't assert a specific value for titleAccurate — it only checks the field
 is defined and boolean, precisely because this is an ambiguous case.

JiwaniZakir:
The distinction between "missing defense" and "demonstrated exploit chain" is the right place to draw the line — the title should reflect what the POC proves, not what an attacker *could theoretically* do with the weakness. One thing worth considering: how does the judge handle cases where the title uses softer language like "may allow" or "could enable"? Those are still inflated if the POC doesn't attempt the full chain, but the signal is weaker and you'd want to avoid flagging findings that are accurately hedged. Also worth defining whether the validator checks only the title or also the summary/impact field, since inflation often bleeds into both.

Yuvanesh-ux:
@JiwaniZakir All right dude no offense, but we've got to at least check the claude output if you copy paste my message into the chat. Pls refrain from AI slop comments.

Yuvanesh-ux:
@cursoragent rebase witht he latest canary branch

PR #568, opened 2026-04-07
fix: auto-disable extended thinking for unsupported models & add Space shortcut hint

jorgeraad:
### What does this PR do?

Fixes the bug where switching to a model that doesn't support extended thinking would hide the toggle UI but leave `reasoningEnabled` set to `true` in state/config. This caused errors when attempting to use the unsupported model because the system would still pass extended thinking options to the provider.

**Root cause:** The UI correctly hid the toggle for unsupported models, but the persisted `reasoningEnabled` flag in config and in-memory state was never cleared when switching models. So the flag was silently carried forward.

**Changes:**

1. **Save-time guard** (`src/tui/context/agent.tsx` — `setModel`): When persisting a model switch, if `reasoningEnabled` is `true` and the new model doesn't support thinking, clear `reasoningEnabled` in both React state and the config file in a single atomic write. This is the primary fix — it ensures stale config can never persist.

2. **Runtime guard** (`src/tui/components/operator-dashboard/index.tsx`): Gate `enableThinking` with `modelSupportsThinking(model.id)` before passing it to the agent. This is defense-in-depth — even if config is somehow stale (e.g. manually edited, race condition), thinking is never requested for an unsupported model at the call site.

3. **`[Space] Extended Thinking` shortcut hint** (`src/tui/components/model-picker/ModelPickerDialog.tsx`): The model picker dialog footer now conditionally shows a `[Space] Extended Thinking` control hint when the selected model supports thinking. Disappears for non-thinking models, matching the checkbox visibility.

**What was considered and rejected:** An initial approach used a `useEffect` to auto-disable `reasoningEnabled` whenever the model changed. This was removed per review feedback — the UI already hides the toggle for unsupported models, so the effect was redundant. The save-time + runtime guard approach is simpler and more targeted.

- All CI checks pass: `bun run tsc`, `bun run lint`, `bun run format:check`, `bun run test` (586 tests pass)
- Manual TUI testing with the following flow:
  1. Selected Claude 3.7 Sonnet (thinking-capable) — footer shows `[Enter] confirm · [Space] Extended Thinking`
  2. Pressed Space to enable extended thinking — checkbox toggles to `[x]`
  3. Navigated to Claude 3.5 Haiku (non-thinking) — footer shows only `[Enter] confirm`, checkbox hidden
  4. Navigated back to Claude 3.7 Sonnet — checkbox shows `[ ]` (unchecked), confirming the save-time guard cleared the state

**Demo video:**

<a href="https://cursor.com/agents/bc-c5b159e5-01d7-44a9-a0ff-1cdd8fb82a21/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fextended_thinking_save_time_guard_demo.mp4">extended_thinking_save_time_guard_demo.mp4</a>

**Key screenshots:**

Thinking-capable model with `[Space]` hint:
Thinking model with Space hint

Extended thinking toggled ON:
Extended thinking toggled on

Non-thinking model — no hint, checkbox hidden:
Non-thinking model without hint

After switching back — save-time guard cleared the state (unchecked):
Save-time guard cleared state

<div><a href="https://cursor.com/agents/bc-c5b159e5-01d7-44a9-a0ff-1cdd8fb82a21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c5b159e5-01d7-44a9-a0ff-1cdd8fb82a21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/core/agents/offSecAgent/tools/documentEndpoint.ts
+++ b/src/core/agents/offSecAgent/tools/documentEndpoint.ts
@@ -208,8 +208,10 @@ Each endpoint creates a JSON file in the assets directory for tracking and analy
         pentestObjectives: input.pentestObjectives,
       };
 
-      const threatModelOutput =
-        await generateThreatModelForEndpoint(ctx, subagentInput);
+      const threatModelOutput = await generateThreatModelForEndpoint(
+        ctx,
+        subagentInput,
+      );
 
       const riskScore = threatModelOutput?.riskScore ?? heuristicRiskScore;
 
--- a/src/tui/components/model-picker/ModelPickerDialog.tsx
+++ b/src/tui/components/model-picker/ModelPickerDialog.tsx
@@ -10,7 +10,9 @@ import { useConfig } from "../../context/config";
 import { useTheme } from "../../theme";
 import { Dialog } from "../../context/dialog";
 import DialogLayout from "../dialog-layout";
+import type { FooterAction } from "../dialog-layout";
 import { ModelPicker } from "./ModelPicker";
+import { modelSupportsThinking } from "../../../core/ai";
 
 interface ModelPickerDialogProps {
   onClose: () => void;
@@ -27,6 +29,15 @@ export default function ModelPickerDialog({ onClose }: ModelPickerDialogProps) {
     setReasoningEnabled,
   } = useAgent();
 
+  const thinkingSupported = modelSupportsThinking(model.id);
+
+  const footerActions: FooterAction[] = [
+    { key: "Enter", label: "confirm", variant: "primary" },
+  ];
+  if (thinkingSupported) {
+    footerActions.push({ key: "Space", label: "Extended Thinking" });
+  }
+
   const title = (
     <text>
       <span fg={colors.primary}>Select AI Model</span>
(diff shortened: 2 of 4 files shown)


jorgeraad:
@cursoragent you added a new effect, but I noticed that before this change, we already had the extended reasoning option only appear when the model selected supported it. This tells me that we likely do not need to introduce a new effect here.

Can you look at this and confirm? If so, please simplify the solution here.

Also this bug tells me that we were persisting to somewhere in state the extended reasoning option and saving it, even if the model did not support that. We should additionally drop this option at save time if the model does not support it. That should be an additional check we do in case there is a bug with the effect and would make it impossible at run time for the extended thinking option to be passed for a model that does not support it.

jorgeraad:
@cursoragent "This branch is out-of-date with the base branch" so please rebase onto the latest changes from canary, typecheck, lint, format, commit, push

kylejryan (approved):

PR #632, opened 2026-04-20
auth: typed errors, 401 retry, real startup preflight

Yuvanesh-ux:
## Summary
Builds on #629 (toast + `/login` rename). Three problems that together caused the "silent session expired after 2 days" bug:

- **Refresh failures were swallowed.** `refreshAccessToken` logged to stderr and returned `null`, which surfaced as a generic `Pensar streaming error (401)`. Now throws `ApexAuthError` with a `reason` discriminator (`refresh_http_error`, `refresh_exception`, `session_dead`, `no_credentials`, `workos_config_unavailable`).
- **The startup preflight was too conservative.** It only fired when there was no refresh token at all — missing the common case where the refresh token *exists* but is already dead on WorkOS's side. Now uses `ensureValidTokenOrThrow` so the refresh path is actually exercised on mount; the existing toast fires on any failure reason other than `no_credentials`.
- **No mid-session retry on 401.** `pensar.ts` now does exactly one silent retry with `{ forceRefresh: true }`. A second 401 throws `ApexAuthError({ reason: 'session_dead', message: 'Pensar session expired. Run /login to reconnect.' })` — a message callers can render verbatim.

Also:
- Concurrent refreshes are serialized (`inFlightRefresh` promise). WorkOS rotates the refresh token on every use, so parallel callers would otherwise race and one would fail with a rotated token.
- Config schema gains `accessTokenIssuedAt` / `refreshTokenIssuedAt`, stamped on login and on every successful refresh.
- Sibling console PR adds distinct `token_expired` vs `token_invalid` 401 codes and a `/api/cli/session` endpoint.

## Test plan
- [ ] `apex login`, verify config gets `accessTokenIssuedAt` + `refreshTokenIssuedAt`.
- [ ] Edit config to set `accessToken` `exp` in the past with valid `refreshToken` → next command silently refreshes, `accessTokenIssuedAt` updates.
- [ ] Replace `refreshToken` with `"invalid"` → TUI startup shows "Your Pensar Console session has expired. Run /login to refresh."
- [ ] Make two simultaneous inference calls right after an `exp` → both succeed (serialized refresh works).
- [ ] Invalidate token server-side mid-session → one silent retry, then typed error with clear message.

## Follow-up (not in this PR)
- WorkOS dashboard refresh-token inactivity timeout — confirm ≥ 7d (ideally 30d). Dashboard-only setting; record decision in `decisions/` afterward.
- Bump `packages/apex` submodule pointer in console repo after this merges.

---

> [!NOTE]
> **Medium Risk**
> Touches WorkOS token refresh and gateway request auth/retry behavior; failures could affect login/session stability and inference requests if error handling or refresh logic regresses.
> 
> **Overview**
> Fixes silent WorkOS session-expiry failures by introducing typed `ApexAuthError` reasons and a new `ensureValidTokenOrThrow` path that surfaces refresh/config problems instead of swallowing them.
> 
> Adds a real startup session preflight in the TUI that actively exercises refresh and warns on any auth failure (except missing credentials), and updates the Pensar gateway client to do a **single 401 retry** that forces token refresh before declaring the session dead.
> 
> Hardens token refresh by serializing concurrent refresh attempts (WorkOS refresh-token rotation) and extends persisted config with `accessTokenIssuedAt`/`refreshTokenIssuedAt`, stamped on login and successful refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70dbe986c257d6a630ce44a7902f6bc7d417a2d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/tui/index.tsx
```diff
@@ -260,27 +260,36 @@ function AppContent({
...
-        isTokenExpired(config.data.accessToken, 60) &&
-        !config.data.refreshToken &&
-        !config.data.pensarAPIKey
-      ) {
-        toast(
-          "Your Pensar Console session has expired. Run /login to refresh.",
-          "warn",
-          8000,
-        );
+    // Preflight the stored session so we catch dead refresh tokens before
```
jorgeraad: There's a number of verbose comments in this PR. Let's cut them down or completely remove any unnecessary ones. For example, in this case, the code speaks for itself and the majority of the comment is just duplicating in English. Extra context that is not captured can be stated briefly.

jorgeraad (approved):
It would really help to have a screen recording or some other verification that this worked as intended.

src/tui/index.tsx
Yuvanesh-ux: @cursoragent Please address the above concern

Yuvanesh-ux:
@cursoragent rebase with the latest canary and babysit the pr until ci checks are finished

Issue #645, opened 2026-04-24
PersistentShell: concurrent execute() calls cross-contaminate stdout/stderr


kylejryan:
## Summary

`PersistentShell.execute()` has no per-shell command queue. When the agent issues parallel `execute_command` calls in a single Claude turn, both invocations share the same `this.current` pointer — last writer wins — and one wrapper's output (stdout, the per-command CUTOVER/EXIT markers, and stderr) ends up accumulated into the *other* command's pending buffers. The result is silent cross-contamination of both stdout and stderr returned to the model.

This is a separate bug from the silent stdout-loss + own-nonce marker leaks fixed in #640. The marker-stripping helper added in #640 cannot fix this case by design: it strips a command's *own* nonce, but the leaked bytes here belong to a *different* command's nonce.

## Forensic evidence (NOCTURNE 004, attack-surface-agent step 4)

Two parallel `execute_command` calls in a single agent step:

| | Call A `toolu_013gEBbqhmGpnRSqMJo9kYCt` | Call B `toolu_011teCwm4UHUs9t4SV1G8E9o` |
|---|---|---|
| description | "Find available wordlists" | "Test Flask-specific paths and more endpoint patterns" |
| command | `find / -name "*.txt" -path "*/wordlist*" …; find …; find …; find …` | `for path in /meridian /auth … /playground; do curl … done` |
| timeout | **10 s** | **120 s** |
| `result.success` | false | false |
| `result.error` | `Command timed out` | `Command timed out` |
| `result.stdout` | `(no output)` ← clean ✓ | `__APEX_8b35a2965745ce81___CUTOVER\n__APEX_8b35a2965745ce81___EXIT_137\n` ← **LEAK** |
| `result.stderr` | `""` | `bash: line 126: 4264 Terminated: 15 find / -name "*.txt" -path "*/wordlist*" …\n4416 Killed: 9   find / -name "*.txt" -path "*/dirb*" …\n4421 Terminated: 15 find / -name "*.txt" -path "*/dirbuster*" …\n5878 Killed: 9   find / -name "*.txt" -path "*/SecLists*" …` |

Call B never spawned a `find`. Yet:
- Call B's stderr contains the SIGTERM/SIGKILL messages for **Call A's** four `find` processes.
[...]

Issue #641, opened 2026-04-24
attack-surface-agent: does not recurse into /debug/* after observing x-debug-mode: enabled signal


kylejryan:
## Summary

The attack-surface-agent observes positive-signal response headers (e.g. `x-debug-mode: enabled`, `x-powered-by`, `server:` banners) but does not translate them into targeted directory enumeration at the signaled prefix. In practice this means that when a target leaks "there is a debug namespace at this host," the agent probes `/debug` as a literal path (gets a 404, which is expected — `/debug` is the namespace root, not an endpoint), then moves on without ever fuzzing `/debug/<WORD>`.

## Where this was discovered

Discovered during failed eval runs against **NOCTURNE** in the evalgate harness (pensar's frontier eval suite for offensive security agents). NOCTURNE-001's primary attack chain requires discovering `/debug/source` on an internal Flask service (Meridian). The service advertises `x-debug-mode: enabled` in its response headers on every request.

Relevant runs:
- Baseline (pre-shell-fix): `01KPXQXR5D7THJJV9EFRSBHF2E` — shell bug made recon impossible, so this behavior was masked.
- Post-shell-fix run: `weave_export_evalgate-apex-nocturne-002_2026-04-24.jsonl` — shell works, agent does real recon with `gobuster`/`ffuf`, and this prompting gap became the dominant failure mode.

## Observed behavior (run 002)

- Agent hit `/health` successfully, saw `x-debug-mode: enabled` in headers.
- Probed `/debug` directly 7 times across different methods → `404` every time (correct; `/debug` is not an endpoint).
- Never executed `gobuster dir -u \$TARGET/debug/ -w <wordlist>` or `ffuf -u \$TARGET/debug/FUZZ -w <wordlist>`.
- Instead fuzzed from the root with various wordlists, none of which contained `source`.
- Stopped at step 39 (~28 min elapsed, 1.79M input tokens) in attack-surface phase, never reaching pentest.

Concrete: the agent saw the `x-debug-mode` signal at step 4 and still only tried `/debug` (literal) and root-level fuzzing through step 39.

## Hypothesis

Either:
1. **Prompting gap** — the attack-surface-agent's system prompt does not list "positive-signal response headers trigger directory recursion at the signaled prefix" as an enumeration heuristic.
2. **Reasoning gap** — the model sees the signal but fails to plan the correct follow-up.

[...]

PR #651, opened 2026-04-27
fix: bundle wordlists, expose as agent capability, and wire env-context across operator + autonomous agents (#642)

### LINKED ISSUE #642: attack-surface-agent: depends on Kali-only wordlist path /usr/share/wordlists/dirb/common.txt
## Summary

The attack-surface-agent defaults to `gobuster -w /usr/share/wordlists/dirb/common.txt` for directory enumeration. This path only exists on Kali Linux and a small number of other security-focused distros. On macOS (operator mode), vanilla Linux, Alpine, and most Docker base images, the file does not exist. When gobuster fails with `wordlist file does not exist`, the agent is forced to build tiny ad-hoc wordlists in its scratchpad, which are far too small to cover realistic endpoint spaces.

## Where this was discovered

Discovered during failed eval runs against **NOCTURNE** in the evalgate harness (pensar's frontier eval suite). The eval runs APEX locally on macOS, so the Kali path was unavailable.

Relevant run: `weave_export_evalgate-apex-nocturne-002_2026-04-24.jsonl`.

## Observed behavior (run 002)

- Step 3: first gobuster run failed immediately:
  ```
  2026/04/24 10:10:06 [ERROR] wordlist file "/usr/share/wordlists/dirb/common.txt" does not exist
  ```
- Steps 4-6: agent retried gobuster with different flag combinations (`-b ""`, `-b "404"`, `--no-status-code`) — none of these fix the missing wordlist.
- Step 9 onward: agent pivoted to building its own wordlists in `scratchpad/wordlist.txt`, `wordlist2.txt`, ..., `wordlist5.txt`, `mega_wordlist.txt`. Typical sizes were tens to low hundreds of entries.
- None of the scratchpad wordlists contained `source`, which is the word needed to discover NOCTURNE's `/debug/source` endpoint.
- Agent spent ~5 steps and a measurable chunk of the 1.79M attack-surface tokens cycling through gobuster flag permutations before giving up on the missing-wordlist error.

## Impact

- Agent is unreliable on non-Kali hosts, which includes the default operator-mode TUI experience on macOS and most CI / Docker deployments.
- Failure is silent-ish: the LLM gets a gobuster error message, interprets it as "tool broken," and pivots to worse alternatives rather than recognizing "I need a different wordlist path."
- Tokens wasted on retries with unchanged-but-different flags.

## Proposed direction

Two tracks, not mutually exclusive:

1. **Bundle a default wordlist with APEX.** Ship e.g. `assets/wordlists/common.txt` (a curated 10-20k entry list — SecLists `common-lowercase.txt` or similar) inside the APEX source tree, reference it by absolute path resolved from `__dirname`. Prompt the agent to default to this path. Works on every platform regardless of host-installed wordlists.
2. **Fail-fast detection.** If gobuster returns the specific "wordlist file does not exist" error, the agent should recognize this as a path problem, not a tool problem, and switch to a known-good path on first retry — not burn 3+ retries with unchanged wordlists.

Bundling is the cleaner fix; fail-fast is a belt-and-braces improvement.

## Bonus

Consider bundling multiple wordlist tiers (tiny / common / large) and teaching the prompt to pick based on elapsed recon time.

## Related

- Companion: #641 (attack-surface-agent doesn't recurse on positive response-header signals) — both reduce the chance of finding `/debug/source` on NOCTURNE.
- Shell-capture PR that unblocked the recon path in the first place.

kylejryan:
## Summary

Closes #642.

Apex agents (operator mode and autonomous sub-agents) reach for `gobuster`/`ffuf`/`dirb`/`wfuzz` and default to `/usr/share/wordlists/dirb/common.txt` from training-data bias. That path only exists on Kali. On macOS, vanilla Linux, Alpine, and most Docker base images the file is missing — `gobuster` errors out, the agent burns retries on flag permutations, then falls back to ~50-entry hand-rolled wordlists in `scratchpad/`. Per NOCTURNE 002, those scratchpad lists missed `source` and the agent never discovered `/debug/source`.

This PR has three parts:

1. **Bundle three wordlist tiers inside the npm-distributed package** so a known-good `-w` argument is always available regardless of host.
2. **Expose them as a first-class agent capability** by injecting an `[BUNDLED ASSETS]` block into the system prompt as an *authoritative inventory* — so when a user asks "do you have wordlists?" the agent answers from context instead of probing the filesystem.
3. **Wire `detectOSAndEnhancePrompt` into operator-mode** which had been silently bypassing it (passing its own `system:` short-circuited the agent constructor's enhancement). Without this, operator agents shipped without `[ENV CONTEXT]` or `[BUNDLED ASSETS]` and had no idea what wordlists or tools were available.

## What's bundled

| File | Entries | Source |
| --- | --- | --- |
| `assets/wordlists/tiny.txt` | 227 | Custom-curated: most-frequently-hit common paths plus a fixed core of high-signal security paths (`.env*`, `.git/config`, `.well-known/security.txt`, `actuator/env`, `swagger.json`, `openapi.json`, `graphql`, `wp-config.php`, etc.). For sub-second smoke probes / time-pressured runs. |
| `assets/wordlists/common.txt` | 4,750 | Verbatim [SecLists `Discovery/Web-Content/common.txt`](https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/common.txt). The default tier. |
| `assets/wordlists/large.txt` | 29,999 | Verbatim [SecLists `Discovery/Web-Content/raft-medium-directories.txt`](https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/raft-medium-directories.txt). Escalation only. |

`assets/wordlists/LICENSE` ships the SecLists MIT license verbatim for attribution. `assets/wordlists/README.md` documents provenance and refresh instructions. `assets/` is added to the `package.json` `files` array so the directory ships with `npm publish`.

## How the agent picks a tier

Driven by an `[BUNDLED ASSETS]` block emitted at runtime by `detectOSAndEnhancePrompt` (in `src/core/agents/specialized/utils.ts`). The block has two distinct jobs:

**1. Capability inventory (introspection).** The block opens with an explicit framing instruction so the agent recognizes it as the authoritative answer to "what wordlists / capabilities do you have?":

> This block is your authoritative inventory of wordlist assets shipped with the CLI. When the user asks what wordlists / assets / capabilities you have, answer directly from the entries below — do NOT probe the filesystem (`ls /usr/share/wordlists`, `which gobuster`, `find / -name wordlists`, etc.). Those paths are not where these live; the inventory is here.

This was added after observing the operator agent run `env | grep wordlist`, `which gobuster`, and `ls /usr/share/wordlists/` in response to capability-discovery questions, ignoring the block entirely. The original "Reference only" wording read to the model as passive config and was discarded; explicit inventory framing flipped the behavior.

**2. Reactive tier-selection policy (tool use).** When the agent has *already decided* to invoke a wordlist-consuming tool, the block tells it which tier to pick. Critically, the language never recommends or implies the agent should run such tools when the task doesn't call for it:

> These paths can be passed as `-w` arguments to gobuster/ffuf/dirb/wfuzz/dirsearch, OR iterated line-by-line in shell loops and `http_request` scripts. Their presence is NOT a reason to run a wordlist-based tool; choose techniques based on the task.
>
> If you do invoke a wordlist-based tool: default to `DEFAULT_WORDLIST`. Use `TINY` only under explicit time pressure or for a first-pass smoke probe. Use `LARGE` only after `DEFAULT` finishes and the target still looks under-mapped, or when the user explicitly asked for a deeper scan. Do NOT chain tiers automatically. Do NOT assume `/usr/share/wordlists/*` exists — it is missing on macOS, Alpine, most Docker images, and CI.

Unit tests assert both the inventory framing and the reactive disclaimer text are present, so future prompt edits can't silently regress to either prescriptive wording or passive-config wording.

### Operator-mode escalation rule

The operator agent gets one extra rule via `src/core/agents/offSecAgent/prompt.ts`:

> Wordlist tools (operator-mode addition). Tier selection is governed by the `[BUNDLED ASSETS]` env block. Operator-only rule: before escalating to `LARGE_WORDLIST`, confirm with the user via the `response` tool with a one-sentence rationale, unless the user pre-directed (e.g. "deep scan", "use the larger wordlist", "quick smoke check").

Sub-agents (`BlackboxAttackSurfaceAgent`, `AuthenticationAgent`) run autonomously with their own pre-enhanced system prompts and don't carry this rule — they decide tier escalation on their own signals. Intentional: a sub-agent that blocks waiting for user input it can't get is worse than one that escalates autonomously.

## Operator-mode wiring fix

While validating end-to-end, I found that the operator dashboard's prompt path was bypassing `detectOSAndEnhancePrompt` entirely. `OperatorDashboard` builds its system prompt via `buildOperatorSystemPrompt` (in `src/tui/components/operator-dashboard/logic.ts`) and passes it as `system:` to `runOffensiveSecurityAgent`. Inside the agent constructor:

```ts
const baseSystemPrompt =
  input.system ??
  detectOSAndEnhancePrompt(buildBaseSystemPrompt({...}));
```

The `??` short-circuits when `input.system` is provided. The operator agent has therefore *never* received `[ENV CONTEXT]` or `[BUNDLED ASSETS]` in its system prompt — from the model's perspective, no one had ever told it the bundled wordlists existed.

Fix: wrap `buildBaseSystemPrompt(...)` with `detectOSAndEnhancePrompt(...)` inside `buildOperatorSystemPrompt`. The base prompt now leads with `[ENV CONTEXT]` → `[BUNDLED ASSETS]` → `# Operator Mode`. Two new regression-guard unit tests in `logic.test.ts` assert both blocks are present.

The autonomous `BlackboxAttackSurfaceAgent` and `AuthenticationAgent` were already wrapping correctly at their call sites and were unaffected.

A wider audit revealed the same gap in `pentest/agent.ts`, `environment/agent.ts`, `patching/agent.ts`, and `whiteboxAttackSurface/agent.ts`. **Filed as #652** to keep this PR focused.

## Path resolution

`src/core/assets/wordlists.ts` exports `getBundledWordlists()`, which walks up from `fileURLToPath(import.meta.url)` looking for the first directory containing both `package.json` (with `name === "@pensar/apex"`) and `assets/wordlists/`. Handles:

- Dev runs from source (`bun run start` → `src/...`)
- Bundled CLI runs (`build/cli.js`)
- Vitest from arbitrary cwds
- Global npm installs (assets sit alongside `build/` under the install root)

Returns `null` if the directory is missing (e.g. a `bun build --compile` binary that didn't bundle the assets), so callers degrade gracefully — no `[BUNDLED ASSETS]` block injected, no false promises to the model.

## Out of scope (filed as follow-ups)

- **#652** — Wire `detectOSAndEnhancePrompt` into the remaining shell-capable subagents (`pentest/agent.ts`, `environment/agent.ts`, `patching/agent.ts`, `whiteboxAttackSurface/agent.ts`). Same one-line fix at each call site; same regression-guard test pattern.
- **#641** — Recursion-on-positive-header-signals for the attack-surface agent.
- **#645** — Concurrent `execute_command` cross-command output bleed in `PersistentShell` (separate failure mode from the now-merged #640 stdout-capture fix).
- **Operator-style escalation gating across the swarm.** Today only the top-level operator agent defers `LARGE_WORDLIST` to the user; spawned sub-agents decide on their own. Gating the entire swarm would require a session-config flag the env-context reads — significant scope expansion.

## Validation

| Check | Result |
| --- | --- |
| `bun run tsc` | Clean |
| `bun run lint` | Clean |
| `bun run format:check` | Clean |
| Targeted tests | 115 passed, 1 skipped (pre-existing macOS `tail -f` skip), 0 failed |
| Full suite | 786 passed, 16 skipped, 2 failed — both `ai.test.ts` failures verified to fail identically on plain canary (require an `ANTHROPIC_API_KEY` for live haiku calls; not an env we set in this shell) |
| Bundle | Rebuilt; `authoritative inventory`, `TINY_WORDLIST`, `BUNDLED ASSETS` confirmed present in `build/` chunks |

The new tests guard:

- All three tier paths are absolute, files exist, sizes are non-zero, `tiny < common < large`
- `tiny.txt` line count stays in the curated 150–300 range (catches drift on upstream re-pull)
- High-signal entries (`.env`, `.git/config`, `.well-known/security.txt`, `swagger.json`, `graphql`, `wp-config.php`, `actuator/env`) are present in `tiny.txt`
- `[BUNDLED ASSETS]` block contains all three `*_WORDLIST=` variables
- `buildOperatorSystemPrompt` output contains `[ENV CONTEXT]` and the three tier variables (operator wiring regression guard)
- The block is framed as an authoritative inventory (`authoritative inventory`, `answer directly from the entries below`)
- Anti-filesystem-probe directive is present (`do NOT probe the filesystem`, `/usr/share/wordlists`, `which gobuster`)
- Broadened-use language is present (`iterated line-by-line`, `http_request`)
- Reactive disclaimer is present (`NOT a reason to run a wordlist-based tool`, `Do NOT chain tiers automatically`)
- `/usr/share/wordlists/*` non-assumption is present
- Operator-mode escalation rules don't leak into the env-context block (those live in the operator prompt only)
- `[BUNDLED ASSETS]` is positioned after `[ENV CONTEXT]` and before the original prompt

## End-to-end demo (the proof-of-life that motivated the rebase)

After rebasing onto canary (which now contains the merged #640 `PersistentShell` stdout-capture fix), I ran the exact scenario that previously failed:

```
> Hallo where ik ur word lists

Hey! Here are the wordlists I have available:
[lists tiny / common / large with paths and sizes — answered from inventory, no filesystem probes]

> Can you run the tiny wordlist on https://coffee-shop-production-db66.up.railway.app/

[runs gobuster dir -u ... -w /Users/.../tiny.txt → returns full Gobuster v3.8.2 banner + 18 lines of real output, finds /admin]
```

Both halves work end-to-end:
- **Capability discovery** (this PR) — agent answers from inventory, lists tiers, surfaces the operator-mode escalation gate, no filesystem probing
- **Wordlist execution** (depends on merged #640 + this PR) — `execute_command` returns real `gobuster` output instead of `(no output)`; tier-selection policy honored (used tiny when asked, offered to escalate to default and asked before going to large)

## Test plan

- [x] Pull this branch on macOS, run the TUI, confirm the env-context block in any agent's system prompt shows absolute paths to `tiny.txt`/`common.txt`/`large.txt`
- [x] Run `gobuster dir -u <local-target> -w "$DEFAULT_WORDLIST"` against a target and verify it succeeds (no "wordlist file does not exist")
- [x] Verify capability-discovery answers from inventory with no filesystem probes
- [x] Verify tier-selection policy honored (used tiny when asked, offered escalation)
- [ ] Re-run NOCTURNE 002 scenario locally; confirm gobuster succeeds on the first attempt and discovers `/debug/source` (deferred — the demo above is a proxy and the rebase off canary unblocks NOCTURNE for whoever picks it up next)
- [ ] Sanity-check the sub-agent path: trigger an autonomous attack-surface run and confirm it picks a tier on its own without trying to call `response`

---

> [!NOTE]
> **Low Risk**
> Low code risk since this change only adds static asset files, but it increases package size and introduces third-party content that must be kept properly attributed/licensed.
> 
> **Overview**
> Adds a new `assets/wordlists/` bundle for the CLI, including SecLists MIT `LICENSE`, a `README.md` documenting tiering/provenance/update steps, and a large `common.txt` directory wordlist intended for tools like `gobuster`/`ffuf` to avoid relying on host-specific paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4c3ad2d7919fdc7010c3927e16244d1e6929523. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kylejryan:
## Follow-up: capability-inventory framing for `[BUNDLED ASSETS]`

Pushed `820f1cfe` after observing a regression in operator-mode dialogue. When asked "do you have access to wordlists?", the agent was running `ls /usr/share/wordlists`, `find / -name wordlists`, `which gobuster` and reporting "no wordlists found" — completely ignoring the `[BUNDLED ASSETS]` block already in its own system prompt.

### Root cause

The block opened with "Reference only" and was framed entirely around *tool selection*. It read as passive config to consult only after the agent had already decided to run a wordlist-based tool — not as a capability inventory to surface when the user asks an introspection question.

### Fix

Reframed the top of the block to make the discovery use case explicit, while preserving the reactive tool-selection guarantee:

- **Inventory framing first.** "This block is your authoritative inventory of wordlist assets shipped with the CLI. When the user asks what wordlists / assets / capabilities you have, answer directly from the entries below."
- **Anti-filesystem-probe directive.** Explicitly tells the agent not to run \`ls /usr/share/wordlists\`, \`which gobuster\`, or \`find / -name wordlists\` — those are not where these live.
- **Broadened use cases.** Wordlists can be passed as \`-w\` args **or** iterated line-by-line in shell loops / `http_request` scripts. The previous wording ("for tools that take a -w wordlist") was unnecessarily narrow.
- **Reactive disclaimer preserved verbatim.** "Their presence is NOT a reason to run a wordlist-based tool; choose techniques based on the task." The tier-selection policy and no-auto-chain rule are unchanged.

### Why prompt-only

`detectOSAndEnhancePrompt` already reads `getBundledWordlists()` at runtime and emits the block dynamically — there's nothing hardcoded to maintain. The architecture goal ("agent inspects itself, no hand-maintained list") is intact; only the wording needed to change so the model recognizes the block as a capability inventory rather than tool metadata.

### Test updates

Replaced assertions on the removed `"Reference only"` string with positive assertions for:
- inventory framing (`authoritative inventory`, `answer directly from the entries below`)
- anti-filesystem-probe directive (`do NOT probe the filesystem`, mention of `/usr/share/wordlists` and `which gobuster`)
- broadened use language (`iterated line-by-line`, `http_request`)

Existing guards on the no-/usr/share/wordlists assumption and operator-mode escalation rules retained.

### Verification

- `bunx tsc --noEmit` clean
- `bun test src/core/agents/specialized/utils.test.ts src/core/assets/wordlists.test.ts` → 17 pass, 0 fail
- `bun run lint` clean
- `bun run format:check` clean

kylejryan:
## Root cause: operator prompt path was bypassing `detectOSAndEnhancePrompt` entirely

After `bun run build` the agent **still** probed the filesystem when asked about wordlists — running `ls /usr/share/wordlists/`, `which gobuster`, `env | grep wordlist`, etc. That ruled out the framing fix as sufficient and pointed at something deeper.

### What was actually happening

`OperatorDashboard` builds its system prompt via `buildOperatorSystemPrompt` in `src/tui/components/operator-dashboard/logic.ts` and passes it as `system:` to `runOffensiveSecurityAgent`:

```ts
agentResult = await runOffensiveSecurityAgent({
  ...commonInput,
  system: systemPrompt,      // ← provided
  session,
});
```

In the `OffensiveSecurityAgent` constructor:

```ts
const baseSystemPrompt =
  input.system ??
  detectOSAndEnhancePrompt(buildBaseSystemPrompt({...}));
```

The `??` short-circuits when `input.system` is provided. **The operator agent has never received `[ENV CONTEXT]` or `[BUNDLED ASSETS]` in its system prompt.** From the model's perspective, no one ever told it the bundled wordlists existed — so its only option was its training-data prior to probe `/usr/share/wordlists`.

The autonomous `BlackboxAttackSurfaceAgent` and `AuthenticationAgent` wrap their own prompts at their call sites, so they were unaffected. Operator was the broken path.

### Fix (`23f61e94`)

Wrap `buildBaseSystemPrompt` with `detectOSAndEnhancePrompt` inside `buildOperatorSystemPrompt`. The operator prompt now correctly leads with the env-context and bundled-assets blocks:

```
[ENV CONTEXT]
OS: darwin | InDocker: no | Kali: no
Tools available: curl, git, gobuster, nc, nikto, nmap, sqlmap, tcpdump
Tools missing: hashcat, hydra, john, socat, tshark, wget
[/ENV CONTEXT]

[BUNDLED ASSETS]
This block is your authoritative inventory of wordlist assets shipped with the CLI...
TINY_WORDLIST=/.../assets/wordlists/tiny.txt (~200 entries)
DEFAULT_WORDLIST=/.../assets/wordlists/common.txt (~4.7k entries)
LARGE_WORDLIST=/.../assets/wordlists/large.txt (~30k entries)
...
[/BUNDLED ASSETS]

# Operator Mode
...
```

### Regression guards

Two new unit tests in `logic.test.ts` pin this down:

```ts
it("includes the [ENV CONTEXT] block from detectOSAndEnhancePrompt", () => { ... });
it("includes the [BUNDLED ASSETS] inventory so the agent can answer capability questions", () => { ... });
```

If a future refactor strips the env-context wrap, these will fail loudly instead of silently regressing the operator's self-introspection. Comment block in the source documents the constructor short-circuit so the next person doesn't reintroduce the bug.

### Wider audit (separate follow-up)

`grep`ing for `system:` across all `OffensiveSecurityAgent` callers reveals the same gap in several specialized sub-agents that have shell access:

| File | Wraps with `detectOSAndEnhancePrompt` |
|------|---------------------------------------|
| `attackSurface/blackboxAgent.ts` | ✅ |
| `authenticationAgent/agent.ts` | ✅ |
| **`pentest/agent.ts`** | ❌ — major code path (`/pentest` workflow) |
| `environment/agent.ts` | ❌ (ironic given its name) |
| `patching/agent.ts` | ❌ |
| `whiteboxAttackSurface/agent.ts` | ❌ |

The judge / scorer / comparison agents (cvss, findingJudge, benchmarkComparison) don't have shell access and don't need it.

I'll file a follow-up issue to clean these up — keeping this PR focused on the operator path you're actively testing.

### Verification

- `bunx tsc --noEmit` clean
- `bun test src/tui/components/operator-dashboard/logic.test.ts src/core/agents/specialized/utils.test.ts src/core/assets/wordlists.test.ts` → 94 pass, 0 fail
- `bun run lint`, `bun run format:check` clean
- `bun run build` succeeds; `grep` confirms `authoritative inventory`, `TINY_WORDLIST`, `BUNDLED ASSETS` are all present in the relevant chunks
- Direct invocation of `buildOperatorSystemPrompt` confirmed the assembled prompt now opens with `[ENV CONTEXT]` → `[BUNDLED ASSETS]` → `# Operator Mode`

kylejryan:
Btw this commit is not this large. It's only +147 -5 (excluding the wordlists, tests, and .md updates)

Yuvanesh-ux (approved):
LGTM

kylejryan:
@Yuvanesh-ux Bet. I'm rebasing then merging 🫡

PR #656, opened 2026-04-27
fix: prevent subagent hub scroll position reset on log updates

jorgeraad:
### What does this PR do?

Fixes the issue where the subagent list view scrolls to the top anytime a new log comes in for a subagent, making it virtually impossible for users to scroll down to older subagents until activity slows down.

**Root cause:** In `SubagentHub`, a `useEffect` called `scrollToIndex` with dependencies `[focusedIndex, sorted]`. Every subagent log update triggers a store update, which produces a new `sessions` Map reference, which in turn creates a new `sorted` array reference via `useMemo`. This caused the scroll effect to re-run on every log event. When `focusedIndex` is 0 (the default), `scrollToIndex` calls `scrollBox.scrollTo(0)`, forcing the list back to the top.

**Fix:** Store `sorted` in a ref (`sortedRef`) and only include `focusedIndex` in the effect's dependency array. The scroll effect now fires only when the user actually navigates (arrow keys / j/k), not when subagent data updates arrive. The ref always points to the latest `sorted` array so `scrollToIndex` still operates on current data when it does fire.

This is a minimal, targeted change — one file modified (`subagent-hub.tsx`), no behavior changes for keyboard navigation.

- ESLint passes (`bun run lint`)
- TypeScript type check passes (`bun run tsc`)
- All unit tests pass (`bun run test` — 762 passed, 15 skipped; 2 pre-existing AI stream test failures that require a live API key)
- Added 7 new unit tests (`subagent-hub.test.ts`) that document the root cause:
  - `sortSessions` order stability (3 tests)
  - Store snapshot identity changes on every log update (4 tests) — proves that `sorted` gets a new reference on every `appendText`, even when the sort order is unchanged, which is why depending on it in the scroll effect caused continuous scroll-to-top resets

<div><a href="https://cursor.com/agents/bc-6806d4f7-3110-4327-9c98-2c2073aea2c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6806d4f7-3110-4327-9c98-2c2073aea2c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


jorgeraad:
https://github.com/user-attachments/assets/0bb251b0-5654-4e9b-b2dd-049796a879e4

KeremP (approved):

KeremP:
just need to rebase

PR #659, opened 2026-04-28
Add screenshot-safe obfuscation mode (`--obfuscate` / `/obfuscate`)

joshkotrous:
### What does this PR do?

Adds a global `--obfuscate` flag (alias `--redact`, `-O`) **and** a runtime `/obfuscate` slash command that put the Pensar TUI into **obfuscation mode**. When active, every layer of the UI passes its text through a pattern-based redaction engine before rendering, so engagement screenshots can be shared without leaking sensitive information.

### Engine — `src/core/obfuscation/`

A pattern-based redactor with a stable, process-lifetime mapping (the same input always produces the same placeholder, so `[EMAIL_1]` is always Alice's email).

| Category | Examples redacted |
| --- | --- |
| `URL`, `HOST` | `https://api.acme-corp.internal/v1/users`, `staging-console.acme.dev` |
| `EMAIL`, `PHONE` | `alice@acme.com`, `+1 415 555 0123` |
| `IPV4`, `IPV6`, `MAC` | `10.0.42.17`, `pentest-10-0-0-1`, `fe80::1`, `00:1A:2B:3C:4D:5E` |
| `UUID` | `550e8400-e29b-41d4-a716-446655440000` |
| `JWT`, `TOKEN` | bearer tokens, `sk-…`, GitHub PATs, AWS access keys, Cognito region IDs (`us-east-1_X8rlYugE7`), opaque 20+ char client IDs |
| `HEX` | long hex blobs |
| `PATH` | `/Users/alice/...`, `C:\Users\alice\...`, `/v1/external/sendgrid/{tenant_id}`, `/__debug/pprof/heap`, `/docs/`, single-segment routes (`/login,`, `/dashboard `, `/leads.`), and template-style relatives (`media/{tenant_id}`) |
| `CARD` | Luhn-validated card numbers |
| `ORG` | `Acme Corp`, `Globex LLC`, `Initech Industries`, plus bare org references via the alias system (see below) |

Notes on detection:

- **No host allowlist.** Every detected hostname is redacted, even apparently public ones (`pensar.dev`, `github.com`, …). The only filter on `HOST` is a source-code filename heuristic so `config.json` and `app.py` aren't mis-redacted.
- **Dashed IPv4** in session names / reverse-DNS labels (`pentest-10-0-0-1`, `ec2-54-12-34-56`) is caught with lookbehind/lookahead guards so longer hyphenated number sequences and date-shaped strings (`2026-04-28-12`) don't false-match.
- **Single-segment slash paths** like `/login,`, `/dashboard ` are redacted only when a non-segment, non-newline character follows, so bare slash commands (`/help` at end-of-line / EOS) stay intact for the command router.
- **Org alias system.** When the engine redacts a host or URL, it learns the registrable label (`acme` from `acme.com`) and stores it as an alias for the same placeholder. A final case-insensitive whole-word pass replaces bare references to the org (`the Acme corporate tenant`, `acme-login-recon`) with the same placeholder, so the org name is consistent across surfaces. A blocklist of generic provider labels (AWS, Cloudflare, GitHub, …) prevents the alias system from ever redacting third-party brand names. An `aliases: false` opt-out is exposed for input round-trip flows that need the literal string back.
- Stop-words (`Pensar`, `OpenAI`, `JSON`, …) prevent ORG false positives on common technical terms.

### Placeholder format

Square brackets: `[EMAIL_1]`, `[HOST_1]`, …. Angle brackets get parsed as inline HTML by `marked` and silently dropped from rendered messages, so brackets are the only form that survives every render path.

### TUI integration — centralized rendering patch

Rather than every component calling `obfuscate(...)` on its own strings, the TUI installs a one-time monkey-patch on the OpenTUI text primitives:

- **`TextNodeRenderable.fromString`** (used by the React reconciler's `createTextInstance` on every JSX text leaf) — redacts on creation.
- **`TextNodeRenderable.prototype.children` setter** (used by `commitTextUpdate` on every text update) — redacts on every change.

Each patched node remembers its original (unobfuscated) text in a symbol-keyed slot. On `/obfuscate` toggle, a tree-walker re-emits every node's children from its stored original through the same patched setter, so already-rendered prose retroactively reflects the new state — without it, React's reconciler would skip `commitTextUpdate` on existing text whose string value hasn't changed and the toggle would only affect newly-committed text.

The walker descends into `TextRenderable.textNode` explicitly because OpenTUI keeps each `<text>`'s text-node subtree on a side-channel — `TextNodeRenderable extends BaseRenderable` (not `Renderable`), so it's invisible to the standard `Renderable.getChildren()` traversal.

### Carve-outs (text that bypasses the patch)

A handful of paths route through `TextRenderable`'s `content` setter or a different primitive entirely, and are obfuscated upstream at the boundary:

- **`markdownToStyledText`** — bakes obfuscation into chunks before they become `StyledText`, since assistant messages render via `<text content={styledText}>`.
- **`result-registry.obfuscateStyledText`** — applied to tool result `styledText` (syntax-highlighted previews) for the same reason.
- **`PromptInput`** — runs `obfuscate()` against the textarea buffer on every content change, replacing recognised values with placeholders as the user types. On submit, paste-extmark expansion runs first, then `deobfuscate()` expands every placeholder back to its original — so the agent always receives real values while screenshots of the prompt show only placeholders. The textarea reacts to live `/obfuscate` toggles, redacting or expanding the current buffer in place. Slash commands are bypassed at this layer so `/help` etc. aren't mangled.
- **`ShiningText`** (pending tool-call shimmer) — obfuscates the input string once at the component boundary because the shimmer renders character-by-character via `<text content={ch}>`.

The markdown renderer also passes through inline and block `html` tokens as literal text, so any future angle-bracketed content the assistant emits stays visible instead of being silently consumed by the lexer.

### `/obfuscate` slash command

`ObfuscationProvider` is stateful, seeded by `--obfuscate` (env `PENSAR_OBFUSCATE=1`) and exposes `setEnabled` / `toggle` / `enabled` / `redact` to the React tree. Engine state and React state are kept in lock-step synchronously so toggling never leaves a frame in an inconsistent state. The `CommandRouter` re-binds handler context on every `execute()` so toggles read live engine state instead of a stale closure from the first render. Subcommands: `on`, `off`, `toggle` (default). A toast confirms the new state.

Help text in `pensar --help` documents the flag; the in-TUI help dialog auto-picks up the new `/obfuscate` command from the registry.

- `bun run tsc` ✅
- `bun run lint` ✅
- `bun run format:check` ✅
- `bun run test src/core/obfuscation` ✅ — 43 cases covering the no-allowlist behaviour, dashed-IP detection, route + relative + template-style path patterns, opaque/AWS-region IDs, ORG/alias logic with provider-blocklist and round-trip opt-out, and `deobfuscate` round-trips.
- `bun run test` ⚠️ — only failures are pre-existing `src/core/ai/ai.test.ts` cases unrelated to this change (retired `claude-3-haiku-20240307` model; identical to `canary`).
- Branch rebased onto latest `canary`. Rebase resolved three conflicts cleanly: footer (kept the new dimensions-aware `showCwd`, dropped the `● obfuscated` indicator), agent-display (`SpinnerDots` → `SpinningDots`), and operator-dashboard header (kept canary's simplified header).
- Manual TUI verification on a real terminal:

https://github.com/user-attachments/assets/d62f9687-d0e8-4570-82e1-5c6eced6fc34

<div><a href="https://cursor.com/agents/bc-f4a58872-ea32-49fb-bf90-5bfbb4183fbf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f4a58872-ea32-49fb-bf90-5bfbb4183fbf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pensar/apex",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "description": "AI-powered penetration testing CLI tool with terminal UI",
   "module": "src/tui/index.tsx",
   "main": "build/cli.js",
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -23,6 +23,18 @@ const args = process.argv.slice(2);
 const command = args[0];
 const version = packageJson.version;
 
+// Detect global --obfuscate flag and propagate to the TUI via env so the
+// flag works regardless of where it appears in argv. The flag is stripped
+// before any per-command parsing so it never collides with subcommand args.
+const OBFUSCATE_FLAGS = new Set(["--obfuscate", "--redact", "-O"]);
+const obfuscateRequested = args.some((a) => OBFUSCATE_FLAGS.has(a));
+if (obfuscateRequested) {
+  process.env.PENSAR_OBFUSCATE = "1";
+  for (let i = args.length - 1; i >= 0; i--) {
+    if (OBFUSCATE_FLAGS.has(args[i]!)) args.splice(i, 1);
+  }
+}
+
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
@@ -122,6 +134,9 @@ threat-model options:
 Global options:
   -h, --help         Show this help message
   -v, --version      Show version number
+  --obfuscate        Run the TUI in obfuscation mode — redacts hostnames,
+                     IPs, UUIDs, emails, paths, tokens, and apparent
+                     company names so screenshots are safe to share.
 `);
 }
 
--- a/src/core/obfuscation/engine.test.ts
+++ b/src/core/obfuscation/engine.test.ts
@@ -67,13 +67,17 @@ describe("obfuscation engine", () => {
 
   it("redacts every hostname including the operator's own infra", () => {
     // No allowlist: pensar.dev, github.com, subdomains thereof are all redacted.
+    // A bare host gets a HOST placeholder; a host followed by a slash-path is
+    // absorbed by the route pattern as a single PATH placeholder, which is
+    // strictly more conservative.
     const out = obfuscate(
       "See staging-console.pensar.dev and github.com/org/repo",
     );
     expect(out).toContain("[HOST_1]");
-    expect(out).toContain("[HOST_2]");
+    expect(out).toContain("[PATH_1]");
     expect(out).not.toContain("pensar.dev");
     expect(out).not.toContain("github.com");
+    expect(out).not.toContain("/org/repo");
   });
 
   it("redacts JWTs and bearer-style tokens", () => {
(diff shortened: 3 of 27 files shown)


joshkotrous:
@cursor rebase this with staging, remove the obfuscated indicator from the footer

PR #663, opened 2026-04-28
fix(persistentShell): serialize execute() with FIFO mutex (#645)

### LINKED ISSUE #645: PersistentShell: concurrent execute() calls cross-contaminate stdout/stderr

kylejryan:
## Summary

Fixes #645. Adds a per-shell FIFO promise-chain mutex around `PersistentShell.execute()` so concurrent calls in the same JS tick can no longer race on `this.current` and cross-contaminate stdout/stderr.

## Root cause

`execute()` runs synchronously up to `return new Promise(...)` and inside that body assigns `this.current = pending` and writes its wrapper to bash's stdin. Two concurrent calls both reach the assignment in the same microtask — last writer wins. Bash drains stdin serially (wrapper A then B) but bytes on both pipes route to whichever pending is currently `this.current`, i.e. pendingB for *both* wrappers' output. PendingA returns ` (no output)`; pendingB leaks A's CUTOVER/EXIT markers (under A's nonce, so the marker-stripper from #640 correctly refuses to touch it) and A's bash job-control stderr.

NOCTURNE 004 step 4 (attack-surface-agent) confirmed the cross-contamination forensically. 84% of ` execute_command` steps in that trace issued 2+ parallel calls — every one a roll of the dice. The marker leak was visible (~1.4% of calls); the stderr leak is silent because there's no marker token on stderr at all.

## Fix

A `writeChain: Promise<void>` field threads a FIFO queue through `execute()`. Each call snapshots the chain tail, installs a new tail, awaits the previous, runs its body, and releases the tail from `pending.resolve` (which already covers all 6 completion paths: happy, timeout, abort, close, cancel, write-error).

Three refinements over the issue's sketch:

- **Re-validate disposal/abort after the queue wait.** A queued call can wait minutes; if the agent aborts mid-queue, it bails when its turn comes without consuming bash time.
- **Defensive try/catch around the post-await body.** If anything throws between `await myTurn` and `pending.resolve` being installed (e.g. `randomBytes` entropy exhaustion), `releaseTurn()` still runs and the queue doesn't deadlock.
- **Timer setup happens after `await myTurn`.** A user-supplied timeout now bounds *execution* time, not *queue time + execution time*.

The `extractFallbackStdout` helper from #640 stays — it's still useful as a safety net for the event-loop-starvation case it was originally designed for, even though cross-nonce contamination is now structurally impossible.

## Behavioral change

Parallel ` execute_command` calls from the agent become **serial** through a shared shell. Invisible to the model (each tool callback still resolves independently). Multi-shell parallelism was considered and rejected because it breaks the `cd`/`export`/alias persistence semantics that motivate `PersistentShell` in the first place.

## Tests

5 new regression tests in `persistentShell.test.ts`:

1. Parallel echoes do not cross-contaminate stdout (no foreign-nonce markers, no foreign output).
2. Parallel calls do not cross-contaminate stderr.
3. Parallel-call timeout isolation — timed-out call's SIGTERM/SIGKILL bash messages and CUTOVER/EXIT_137 markers do not leak to the surviving call.
4. FIFO queue ordering preserved across `Promise.all([echo 1, echo 2, echo 3])`.
5. Aborted-while-queued bails when its turn comes without running its bash command (validates refinement (a)).

All 14 pre-existing tests in the file still pass (26 passed, 1 skipped — the macOS-skipped streaming test).

## Verification

- `bun run test src/core/agents/offSecAgent/tools/persistentShell.test.ts` — 26 passed, 1 skipped.
- `bun run lint` — clean.
- `bun run format:check` — clean.
- `bun run tsc` — clean.
- Full `bun run test` — 791 passed, 16 skipped, 2 pre-existing AI-API-key-required failures unrelated to this change (also failing on canary).

## Test plan (post-merge / pre-merge human-driven)

- [ ] Manual TUI smoke: run a few ` execute_command` calls including a forced parallel batch, confirm normal behavior.
- [ ] NOCTURNE re-run: verify zero `__APEX_<nonce>__` strings in any tool result across all parallel-call steps. This is the durable acceptance criterion that a future regression would trip on.
- [ ] Spot-check stderr on a parallel-timeout step: confirm the surviving call's stderr does not contain ` Terminated` / `Killed` messages from the sibling.

## Related

- #640 — silent stdout-loss + own-nonce marker leak fix (lands first; covers the single-command event-loop-starvation case). #645 covers cross-command leaks. Both layers are needed.
- #636 — earlier `cat`-only fallback for missing `stdbuf` (predecessor work).
- #625 — stdin-hijack fix (sibling `PersistentShell` correctness work).

## Forensic citation

NOCTURNE 004, attack-surface-agent step 4: call A `toolu_013gEBbqhmGpnRSqMJo9kYCt` (find wordlists, 10s timeout), call B `toolu_011teCwm4UHUs9t4SV1G8E9o` (curl Flask paths, 120s timeout). B's stdout contained A's CUTOVER + EXIT_137 markers under A's nonce; B's stderr contained the SIGTERM/SIGKILL messages for A's four `find` processes (B never spawned a `find`).

---

> [!NOTE]
> **Medium Risk**
> Introduces a new queuing/mutex mechanism in `PersistentShell.execute()` that changes concurrency behavior and affects command timing/abort/timeout semantics, so regressions could impact agent command execution under load.
> 
> **Overview**
> Fixes a race in `PersistentShell.execute()` where concurrent calls could overwrite `this.current` and cause **stdout/stderr/marker leakage between commands**.
> 
> Adds a per-shell **FIFO promise-chain mutex** (`writeChain` + `acquireTurn`) so parallel `execute()` calls are serialized; queued calls re-check `disposed`/`abortSignal` after waiting, and timeouts now measure *execution time* rather than queue+execution.
> 
> Expands `persistentShell.test.ts` with new parallelism regression coverage (stdout/stderr isolation, timeout isolation, FIFO ordering, and abort-while-queued), while keeping existing long-running stability tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05ed022d9451a5ecfb607373bf6ee8dc6e0d2df9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Post-fix verification: NOCTURNE 005

Re-ran NOCTURNE on `fix/645-persistent-shell-serialize-execute`. The race condition was actually exercised by real agent traffic, and the durable acceptance criteria all hold.

### Acceptance criteria

| Criterion                                                                         | Result                                                  |
| --------------------------------------------------------------------------------- | ------------------------------------------------------- |
| Zero `__APEX_<nonce>__` strings in any tool result                                | **0 hits** across all 54 `execute_command` observations |
| Zero `EXIT_137` / `EXIT_143` markers (the specific shapes leaked in NOCTURNE 004) | **0 hits**                                              |
| Zero cross-stderr contamination from sibling kill messages                        | **0 hits** (1 self-kill, correctly attributed — below)  |

### Was the race actually exercised?

Yes. Concurrent calls were issued 16 times, including 3 triple-parallel steps:

```
total step records            = 70
steps containing exec_cmd     = 35
total exec_cmd observations   = 54
parallel-call distribution    = [(0, 35), (1, 19), (2, 13), (3, 3)]
steps with 2+ parallel exec   = 16  (46% of exec-steps)
max parallelism observed      = 3
```

Compared to the NOCTURNE 004 baseline cited in the issue (84% parallel, 1.4% visible marker-leak rate, plus systemic silent stderr contamination), this run is lower-density on parallelism but still exercised the contended code path 16 times. Zero leaks across all 16.

### The one bash kill-message in the trace is self-attribution

Step 14, `attack-surface-agent`, n=2 parallel:

- **Call A** (`toolu_01FRVhH2JAfZFR6EMpU3faZm`, curl method matrix on `/artifacts`): success, 4023 chars of clean stdout, **empty stderr**, no markers.
- **Call B** (`toolu_01TfKCE9t5yAMaJSV3dpyMLN`, `gobuster dir`): timed out at the 180s wrapper timeout. Its stderr contains:
  ```
  bash: line 537: 90229 Terminated: 15  gobuster dir -u "https://..." ...
  ```
  PID 90229 is running the same `gobuster` command whose text is in Call B's own `command` field — bash's job-control notice for **Call B's own process being killed because Call B itself timed out**. Pre-fix, this message would have race-routed to whichever pending was `this.current` when bash printed it; post-fix it's structurally pinned to the call that owns the killed process. Call A correctly received nothing.

  (Self-stderr-on-self-timeout is a separate, much smaller cosmetic issue — bash job-control prints to its real stderr pipe, bypassing the wrapper's `2>"$__APEX_ERR"` redirection. Not what #645 was about; out of scope for this PR.)

### Other failures are all explainable

5 non-success `execute_command` results across 54 calls (9.3%):

- 3 **scope violations** — agent tried to access disallowed targets; the policy layer rejected the command before it ran. Empty stdout is by design.
- 1 **self-timeout** (the gobuster above).
- 1 **exit code 1** — generic non-zero exit, no contamination signature.

None are shell-level bugs.

### Analysis caveats

- Trace observations store a 2000-char `outputPreview` (16 of 54 results were longer and got truncated). Marker leaks always appear at the *start* of the result (the wrapper emits CUTOVER first, and `extractFallbackStdout` returns them at the beginning when it can't strip), so the preview catches every known leak shape.
- Lower parallelism density than NOCTURNE 004 (46% vs 84%). Worth re-checking on any higher-density trace from a follow-up run, but this run is a strong positive signal on its own.
- No `extractFallbackStdout` fallback paths fired in this trace — agent saw clean output across the board.

**Net:** zero markers, zero EXIT_137/143, zero cross-stderr contamination across 16 parallel-call steps. Acceptance criteria met. Marking ready for review.

### DIFF
--- a/src/core/agents/offSecAgent/tools/persistentShell.test.ts
+++ b/src/core/agents/offSecAgent/tools/persistentShell.test.ts
@@ -7,17 +7,6 @@ const HAS_STDBUF =
   process.platform !== "win32" &&
   spawnSync("stdbuf", ["--version"], { stdio: "ignore" }).status === 0;
 
-/**
- * These tests document and reproduce the "shell eventually stops responding"
- * class of bugs in PersistentShell. They are written as hang-repros: each test
- * exercises a scenario a long-running pentest agent realistically hits, and
- * asserts the contract the tool advertises.
- *
- * Several of these are EXPECTED TO FAIL against the current implementation.
- * They are intentionally left failing so the bugs cannot be silently
- * re-introduced once fixed. They use `it.fails(...)` so the suite stays green
- * while clearly flagging the known-broken behavior.
- */
 describe("PersistentShell — long-running stability", () => {
   const shells: PersistentShell[] = [];
   const make = () => {
@@ -47,16 +36,6 @@ describe("PersistentShell — long-running stability", () => {
     expect(r2.stdout).toContain("two");
   });
 
-  /**
-   * Regression: on macOS without GNU `stdbuf`, the old wrapper relied on
-   * plain `tail -f` for streaming. Because `tail`'s stdout was a pipe it
-   * block-buffered, so short outputs were killed with tail before ever
-   * flushing — every `ls -la`-style command came back with empty stdout
-   * and the agent saw `(no output)`. The current wrapper always runs an
-   * authoritative post-kill `cat` of the per-command tempfile through
-   * bash's real stdout pipe, so `tail` buffering can no longer cause
-   * stdout loss.
-   */
   it("captures short stdout (no-stdbuf safe)", async () => {
     const shell = make();
     const r = await shell.execute("printf 'hello\\nworld\\n'", 5);
@@ -65,21 +44,6 @@ describe("PersistentShell — long-running stability", () => {
     expect(r.stdout).toContain("world");
   });
 
-  /**
-   * Regression: on macOS with SIP, `stdbuf -oL` is a silent no-op on
-   * /usr/bin/tail (DYLD_INSERT_LIBRARIES is stripped from protected
-   * binaries). If the wrapper relies on tail alone to stream stdout,
-   * tail block-buffers (~8 KiB default) and the buffer is discarded
-   * when we kill tail 100 ms after the command finishes — every short
-   * command returns empty stdout. Executed by APEX TUI in operator mode,
-   * by the CLI pentest entrypoint, and by Evalgate-driven runs (e.g.
-   * NOCTURNE), producing `(no output)` for trivial commands and
-   * dominating eval signal with shell brokenness rather than agent
-   * capability.
-   *
-   * These assertions fail on macOS against the pre-patch wrapper and
-   * pass once the authoritative post-kill `cat` phase is in place.
-   */
   it("captures trivial stdout even when tail block-buffers", async () => {
     const shell = make();
     const echo = await shell.execute("echo hello", 5);
@@ -95,20 +59,9 @@ describe("PersistentShell — long-running stability", () => {
     expect(seq.stdout.trim()).toBe("1\n2\n3\n4\n5");
   });
 
-  /**
-   * Regression: tail's default block-buffer is ~8 KiB. A command whose
-   * stdout is just under, at, and just over that boundary must come back
-   * fully intact, with no silent truncation regardless of whether tail
-   * ever flushed its buffer. Exercises the authoritative `cat` phase on
-   * a realistic recon-output size.
-   */
   it("captures output across the 8 KiB block-buffer boundary", async () => {
     const shell = make();
-    // Emit exactly 9000 'x' bytes followed by a newline. Uses /dev/zero
-    // to avoid bash/yes-pipeline byte-counting footguns (e.g.
-    // `yes x | head -c 9000` includes the newlines that `yes` emits, so
-    // stripping them halves the payload). 9000 > 8192 to ensure we
-    // exceed tail's default ~8 KiB block-buffer size.
+    // 9000 > 8192 to exceed tail's default block-buffer.
     const r = await shell.execute(
       "head -c 9000 /dev/zero | tr '\\0' 'x'; echo",
       5,
@@ -117,58 +70,26 @@ describe("PersistentShell — long-running stability", () => {
     expect(r.stdout.replace(/\n$/, "").length).toBe(9000);
   });
 
-  /**
-   * Repro #1 — Stdin hijack by a child process.
-   *
-   * The shell's stdin pipe is shared with every child bash spawns. A command
-   * that reads from stdin (cat/ssh/sudo/python/mysql/read/...) used to
-   * consume bytes meant for bash's own command parsing, wedging the shell
-   * for every subsequent call.
-   *
-   * The fix redirects each user command's stdin from /dev/null, so stdin
-   * readers see EOF immediately and cannot hijack our command pipe. We
-   * verify:
-   *   1. `cat` does NOT hang — it EOFs cleanly (exit 0) in well under the
-   *      timeout, proving stdin isolation works.
-   *   2. The next command still runs immediately afterward.
-   *
-   * Agents hit this constantly in practice because models run things like
-   * `sudo ...` or `ssh ...` mid-pentest.
-   */
   it("stays responsive after a command that normally reads stdin", async () => {
     const shell = make();
 
     const warm = await shell.execute("echo warm", 5);
     expect(warm.exitCode).toBe(0);
 
-    // `cat` with no args normally blocks on stdin. With /dev/null
-    // redirection it EOFs instantly.
+    // `cat` with no args normally blocks on stdin. With /dev/null it EOFs.
     const start = Date.now();
     const hung = await shell.execute("cat", 2);
     const elapsed = Date.now() - start;
     expect(hung.exitCode).toBe(0);
     expect(elapsed).toBeLessThan(1_500);
 
-    // Subsequent command should return within a normal timeout.
     const after = await shell.execute("echo after-cat", 5);
     expect(after.exitCode).toBe(0);
     expect(after.stdout).toContain("after-cat");
   }, 20_000);
 
-  /**
-   * Repro #2 — Process-group leak after timeout.
-   *
-   * `pkill -TERM -P <bashpid>` only reaps direct children of bash. For a
-   * pipeline (`a | b | c`) the workers are grandchildren of bash (children
-   * of the subshell), so they survive the timeout, become orphans, and keep
-   * their inherited fd 1 — feeding Repro #3 below.
-   *
-   * After the timeout, we should see no descendant processes still holding
-   * the shell's stdout pipe.
-   */
   it("cleans up pipeline grandchildren when a command times out", async () => {
     const shell = make();
-    // Prime the shell so we have a stable pid to probe.
     await shell.execute("echo prime", 5);
     const bashPid = (shell as unknown as { proc: { pid: number } }).proc.pid;
     expect(bashPid).toBeGreaterThan(0);
@@ -177,7 +98,6 @@ describe("PersistentShell — long-running stability", () => {
     const timed = await shell.execute("sleep 30 | cat", 2);
     expect(timed.exitCode).toBe(124);
 
-    // Give pkill + the 1s drain window a moment.
     await new Promise((r) => setTimeout(r, 1_500));
 
     const { spawnSync } = await import("child_process");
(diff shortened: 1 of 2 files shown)


jorgeraad (approved):
Nit: can probably cut down on the comments a bit. Don't want to confuse the agents, especially if the comments get stale.

PR #672, opened 2026-04-29
Add new model gpt-5.5-2026-04-23

harshakshit:
This pull request adds support for a new OpenAI model to the list of available models. The new model, `gpt-5.5-2026-04-23`, is now included with its corresponding metadata.

OpenAI model updates:

* Added a new model entry for `gpt-5.5-2026-04-23` with a context length of 200,000 to the `OPENAI_MODELS` array in `openai.ts`.


KeremP:
we have a generate model script in the `scripts` directory for providers like openai, anthropic, etc. that is called during build to populate these :) if you add this to that script we can merge in.

PR #681, opened 2026-04-29
fix: make syntax highlighting theme-aware with per-theme IDE palettes

jorgeraad:
### What does this PR do?

Makes code output syntax highlighting fully **theme-aware**, fixes a series of related light-mode readability bugs, and tightens the scrollbar UX in scrollable surfaces.

**Theme-aware syntax highlighting (original scope):**
- Each of the 18 built-in themes now defines its own syntax palette with dark and light variants, sourced from the canonical IDE color scheme for that theme. Previously, syntax highlighting used a single hardcoded palette for all themes.
- Added 9 new semantic color tokens to `ThemeColors`: `syntaxKeyword`, `syntaxString`, `syntaxComment`, `syntaxNumber`, `syntaxFunction`, `syntaxType`, `syntaxTag`, `syntaxAttr`, `syntaxPunctuation`.
- `highlightCode()` now takes resolved `ThemeColors` and builds the hljs class map from the active theme. `getResultSummary()` and call sites (`tool-renderer.tsx`, `tool-message.tsx`) pass `colors` through. Web search result styling uses theme tokens instead of a hardcoded `WS_COLORS` map.

**Light-mode rendering bugs (uncovered while testing the above):**

1. **Markdown prose rendered as opentui white default in light themes.** `markdownToStyledText` pushed plain-text, br, list-marker, paragraph/heading newline, html, blockquote-body, and space chunks with no `fg`. Opentui's StyledText chunks without `fg` fall back to a hardcoded white default — invisible on light dialog backgrounds — regardless of the parent `<text fg=…>`. Resolve `textColor` from `colors.text` and set `fg` on every chunk that doesn't already carry an explicit color (codespans, links, code blocks, blockquote indicator keep their existing colors).
2. **Unclassed code tokens rendered white in light themes.** Same chunk-fg-undefined issue at a different layer: `parseHljsHtml` pushed text-node chunks with `fg: colorStack[…]` where the initial stack was `[undefined]`. Any token outside an hljs span — whitespace, punctuation like `{ } | ; ,`, plain identifiers like `requireAdmin` that hljs doesn't classify — got `fg: undefined` and rendered as white. Pass `colors.text` as `defaultColor` and seed the stack with it.
3. **Many syntax tokens fell below readable contrast on the dialog background.** Audited all 18 themes × 9 syntax tokens × 2 modes (162 tokens) against `backgroundElement`. 75 tokens (mostly in light mode) were below WCAG AA Large (3:1), some under 2:1. For each failing token, adjusted lightness in HSL space (preserving hue and saturation) until contrast cleared the threshold — 3.0:1 floor for active tokens, 2.5:1 floor for comments (kept intentionally subdued). Also rebuilt the apex theme's syntax palette with hue-stable HSL values that boost saturation ~50% over the prior tone-down so it reads as "muted but not dull". After this pass, 0/162 tokens fail their threshold.

**Scrollbar UX (regression from the convention introduced in #682):**

4. **`scrollbarOptions: { visible: true }` painted a stray solid block on short surfaces.** With nothing to scroll, the thumb spanned the entire track and read as a colored rectangle on the right edge. Dropped `visible: true` on MessageList, SubagentHub, HelpDialog, SessionsDisplay so opentui's auto-visibility hides the bar when there's no overflow.
5. **Thumb painted in `colors.primary` was too loud.** Primary is the brand accent and shouldn't dominate every scrollable surface. Switched all 10 `foregroundColor: colors.primary` track options to `colors.textMuted` so the thumb is a subtle theme-aware gray. Track stays `colors.backgroundElement` so it blends with the dialog body. Updated `AGENTS.md` so the convention reflects the new behavior.

- `bun run tsc` — type check passes
- `bun run lint` — no lint errors
- `bun run format:check` — formatting correct
- `bun run test` — 829 passed, 18 skipped
- Manual TUI verification across light and dark modes on tokyonight, github, dracula, apex, and others

https://github.com/user-attachments/assets/759684fe-f03c-41a5-b2ca-85250eefef75

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,7 @@ On first launch, the TUI shows a "Responsible Use Disclosure" screen that must b
 
 **No state toggling in streaming callbacks.** Avoid calling `setState` to toggle visual indicators (like "Thinking...") directly in high-frequency streaming callbacks (`onTextDelta`, `onToolCallStreaming`). This causes per-frame flicker. Instead, derive display state from the underlying data model (e.g., check if the last message is a tool result rather than toggling a boolean).
 
-**Scrollbar visibility.** Scrollable content (lists, logs, long text) must have `scrollbarOptions={{ visible: true }}` so users can see their scroll position. Never hide scrollbars in scrollable regions.
+**Scrollbars.** Don't pass `scrollbarOptions={{ visible: true }}` — opentui auto-shows the bar when content overflows the viewport and hides it otherwise. Forcing `visible: true` paints a full-track thumb when content fits, which reads as a stray colored block. Style with `trackOptions: { foregroundColor: colors.textMuted, backgroundColor: colors.backgroundElement }` so the thumb stays subtle and theme-aware (don't use `colors.primary` — too loud). Never explicitly hide scrollbars in scrollable regions.
 
 ### Code hygiene
 
--- a/src/tui/components/agent-display.tsx
+++ b/src/tui/components/agent-display.tsx
@@ -159,7 +159,7 @@ export default function AgentDisplay({
         },
         scrollbarOptions: {
           trackOptions: {
-            foregroundColor: colors.primary,
+            foregroundColor: colors.textMuted,
             backgroundColor: colors.backgroundElement,
           },
         },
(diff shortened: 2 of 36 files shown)


jorgeraad:
@cursoragent is it possible to have the code highlighting be theme aware? Basically the colors that would be applied if the theme would be selected in an IDE. There should be dark and light variants for each theme.

jorgeraad:
Fixed it up

https://github.com/user-attachments/assets/759684fe-f03c-41a5-b2ca-85250eefef75

KeremP (approved):

Issue #643, opened 2026-04-24
attack-surface-agent: 1.79M input tokens spent in recon without advancing to pentest phase


kylejryan:
## Summary

On long-horizon targets, the attack-surface-agent can burn disproportionate token budget re-running variations of the same recon tool (different gobuster flags, different ffuf wordlists, different HTTP methods) without meaningfully expanding the attack-surface report or handing off to pentest. The run never reaches pentest-agent, so the full chain of downstream tools (SSRF / authz / SQLi / etc.) never gets exercised.

## Where this was discovered

Discovered during failed eval runs against **NOCTURNE** in the evalgate harness.

Relevant run: `weave_export_evalgate-apex-nocturne-002_2026-04-24.jsonl`.

## Observed behavior (run 002)

- 40 steps over 28.2 min of wall time.
- All 40 steps are `attack-surface-agent`. No `pentest-agent-1` or `pentest-agent-2` records in the export at all.
- Cumulative usage at step 39:
  - inputTokens: **1,788,119**
  - outputTokens: 25,421
  - cacheReadTokens: 0
  - cacheWriteTokens: 75,744
- Input:output ratio ~70:1 — classic sign of the agent re-reading large observation history every step without advancing.
- Tool-call distribution over the run:
  - `execute_command`: 85 (mostly gobuster/ffuf variations)
  - `browser_*`: 9
  - `document_app`: 1
- Agent ran `gobuster` 5×, `ffuf` 10× with different wordlists, flag sets, methods, and URL prefixes. Net new discovered endpoints over all 15 fuzzer runs: **1** (`/health`, which was already known from step 1).
- Final step text was "Now let me document the application and all discovered endpoints" — indicating the agent was about to terminate attack-surface with essentially no new findings beyond what step 1 produced.

## Why this matters

- Attack-surface is the first phase of an APEX pentest workflow. If it never terminates with a usable report, pentest never runs.
- Token cost is large per run. At current usage 1.8M input tokens is a meaningful fraction of the 5M NOCTURNE budget, spent entirely on recon that produced no new information.
- This pattern is not NOCTURNE-specific — any target with a sparse attack surface will trigger the same behavior.

## Hypotheses

1. **No exploration budget cap.** The agent has no explicit "stop enumerating after N failed wordlist attempts" heuristic. It keeps trying because the loop isn't structurally bounded.
[...]

PR #712, opened 2026-05-05
fix(eventBus): forward trace-record across attachChild boundary (#707)

### LINKED ISSUE #707: AgentEventBus: subagent trace-record events not forwarded at attachChild boundary, breaking W&B subagent observability
## Summary

`AgentEventBus.attachChild` forwards a hand-curated allowlist of events from a
child bus to its parent, but `trace-record` is missing from that allowlist. As
a result, every trace record (`step`, `init`, `checkpoint`, `task`) emitted
from any subagent that uses `attachChild` dies on the child bus and never
reaches the W&B Weave uploader (or the operator-dashboard early-buffer
handler) attached on the parent bus.

The local `trace.jsonl` files on disk are unaffected — `appendRecord` writes
to disk before emitting to the bus, in independent try/catch blocks — so the
agent loop continues and produces complete on-disk traces even though no
`trace-record` events reach the parent bus.

## Where this was discovered

Discovered while analyzing `weave_export_nocturne-v2-001_2026-05-04.jsonl`
(NOCTURNE-v2 eval suite, run 001). The Weave UI showed the run as "stalled"
with the last record at `2026-05-04T18:13:27Z`, but the local session
directory had a 318KB `threat-model-meridian_api_service-_health.trace.jsonl`
under `~/.pensar/sessions/<sid>/subagents/` that was actively growing 27
minutes after the last W&B record.

## Observed behavior

In the affected run:
- `attack-surface-agent` produced 19 step records + 1 init record. All
  reached W&B because evalgate's harness invokes `BlackboxAttackSurfaceAgent`
  via the public API path (`src/core/api/attackSurface.ts:50`), which passes
  `eventBus` straight through — no `attachChild`, no child bus.
- The attack-surface-agent's last action (`document_app`) triggered
  `document_endpoint`, which calls `generateThreatModelForEndpoint`. That
  function spawns a `CodeAgent` on a `localBus` wrapped via
  `AgentEventBus.attachChild(localBus, ctx.eventBus, subagentId)`
  (`src/core/agents/offSecAgent/tools/threatModelGenerator.ts:302-303`).
- The `CodeAgent`'s `StepTraceWriter` emits `trace-record` on `localBus`.
  `attachChild` only forwards events listed in `CHILD_BUS_FORWARDED_EVENTS`
  (`src/core/eventBus.ts:75-86`); `trace-record` is not there, so events
  fire into a void.
- Compounding: `generateThreatModelForEndpoint` does
  `await agent.consume()`, so the parent's tool loop is parked while the
  subagent runs. The parent emits no further `trace-record` events either,
  and W&B looks completely frozen.

## Root cause

`src/core/eventBus.ts:75-86`:

```ts
const CHILD_BUS_FORWARDED_EVENTS = [
  "text-delta",
  "tool-call-start",
  "tool-call-delta",
  "tool-call-complete",
  "tool-result",
  "subagent-spawn",
  "subagent-complete",
  "command-output",
  "error",
  "step-finish",
] as const satisfies readonly (keyof AgentEventMap)[];
```

`trace-record` is in the `AgentEventMap` (line 65-68) but not in this
forwarding allowlist. The allowlist is iterated in `attachChild` to register
listeners on the child bus that re-emit on the parent — events not listed
get no listener, so they vanish.

## How the allowlist evolved

Useful context for the long-term fix below.

- **2026-04-06** (commit `4a54732f`, "feat: add child() method to
  AgentEventBus for automatic subagent event forwarding"): Original
  `parentBus.child(subagentId)` forwarder used an exhaustive allowlist
  `AGENT_EVENT_NAMES`, compile-time-asserted via
  `ExhaustiveEventNames<typeof AGENT_EVENT_NAMES>`. Every event in
  `AgentEventMap` was forwarded by construction.
- **2026-04-16** (#614, commit `1eea7e24`): Forwarder refactored to the
  current `attachChild` static + the hand-curated
  `CHILD_BUS_FORWARDED_EVENTS`. The new shape solves a real subagent
  log-duplication issue in the operator dashboard. The hand-curated
  allowlist does not include `trace-record`, which is the proximate cause
  of this issue.

The W&B integration (#517) lands earlier (2026-03-31), but its consumer
registers on the parent bus and so depends on whatever the forwarder is
doing at that boundary.

## Blast radius

`attachChild` is called from 5 sites; all are affected:

| File | Subagent ID pattern |
|---|---|
| `src/core/agents/offSecAgent/tools/threatModelGenerator.ts:303` | `threat-model-<app>-<route>` |
| `src/core/agents/offSecAgent/tools/delegateAuth.ts:302` | `auth-subagent` |
| `src/core/agents/offSecAgent/tools/runAttackSurface.ts:71` | `attack-surface-agent` (whitebox path) |
| `src/core/agents/offSecAgent/tools/runAttackSurface.ts:139` | `attack-surface-agent` (orchestrator-driven blackbox path) |
| `src/core/agents/offSecAgent/tools/spawnCodingAgent.ts:169` | `coding-agent` |

The headless public-API path (`src/core/api/attackSurface.ts:50`) bypasses
`attachChild`, which is why nocturne runs still see attack-surface traces in
W&B. Any orchestrator-driven flow (e.g. `/pentest`, `/operator`) is far more
exposed because it goes through `runAttackSurface` and then spawns multiple
pentest coding-agents via `spawnCodingAgent`.

Two `trace-record` consumers are affected today:
- `src/core/integrations/wandb/upload.ts:70` — W&B Weave uploader.
- `src/tui/components/operator-dashboard/index.tsx:1211` — operator
  dashboard's early-buffer handler that replays into the W&B uploader once
  it attaches.

## Coverage gap

The existing pipeline test
(`src/core/integrations/wandb/upload.test.ts:148`, "EventBus receives trace
records from all agent types") wires multiple writers to a single shared
bus. That covers the headless / public-API topology but not the
`attachChild` topology that orchestrator-driven runs use. The regression
test below exercises the missing case.

## Reproduction

1. Run any flow that spawns a subagent via `attachChild` (simplest:
   `/operator` mode that triggers `document_endpoint`, which spawns a
   threat-model subagent).
2. Configure `WANDB_API_KEY` + `WANDB_ENTITY` so `attachWandbToEventBus`
   activates.
3. Observe in the W&B UI: parent agent records appear; subagent records do
   not.
4. Confirm the subagent is alive and writing locally:
   `ls -la ~/.pensar/sessions/<sid>/subagents/*.trace.jsonl` — the file size
   keeps growing.

## Proposed fix

Add `"trace-record"` to `CHILD_BUS_FORWARDED_EVENTS` in
`src/core/eventBus.ts`. The existing forwarder already injects `subagentId`
when the payload doesn't carry one, but the writer self-tags every record
with `subagentId: this.agentId` (`src/core/agents/offSecAgent/trace.ts:586-589`),
so the records will arrive on the parent bus with the correct
`subagentId` (`threat-model-...`, `coding-agent-...`, etc.) for W&B grouping.

A more robust long-term fix is to revive the type-system guarantee from the
April 6 implementation: make the forwarding allowlist exhaustive over
`AgentEventMap` by default (or at minimum, decide explicitly per-event
whether it's bus-internal vs forwardable), so adding a new event to the map
without forwarding it becomes a compile error. The April 6 commit's
`ExhaustiveEventNames<>` pattern is the precedent.

## Regression test

Add a test in `src/core/eventBus.test.ts` (or `upload.test.ts`) that:
1. Creates a parent bus and subscribes a `trace-record` handler.
2. Creates a child bus via `AgentEventBus.attachChild(child, parent, "sub-1")`.
3. Wires a `StepTraceWriter` to the child bus and calls `recordStep`.
4. Asserts the parent's `trace-record` handler received the event with
   `subagentId === "sub-1"`.

This test would have caught both the original missed forwarding entry and
any future removal of `trace-record` from the allowlist.

## Workarounds (until fixed)

- The local jsonl is the source of truth. Live-tail subagent traces with:
  `tail -f ~/.pensar/sessions/<sid>/subagents/<subagent-id>.trace.jsonl`.
- For post-hoc analysis, parse all `~/.pensar/sessions/<sid>/subagents/*.trace.jsonl`
  files together — they form the complete trace that *should* have appeared
  in W&B.

## Related

- #517 — W&B Weave Integration for Fine-Tuning (where the parent-bus
  `trace-record` consumer lives).
- #614 — Fix subagent log duplication in operator mode main view (where
  the current `attachChild` forwarder shape was introduced).
- #641, #643 — companion NOCTURNE issues. #643's reported "1.79M input
  tokens" may have under-counted subagent token usage if any threat-model
  subagents ran during recon, since their tokens never reached W&B's
  `cumulativeUsage` aggregation. Worth re-baselining once this is fixed.

## Environment

- Apex: `1.8.0` (canary HEAD `147b4685`)
- Eval harness: evalgate / NOCTURNE-v2-001
- Provider: Anthropic (`claude-opus-4-6`)
- Affected since: PR #614 merged 2026-04-16

kylejryan:
Closes #707.

## Summary

`AgentEventBus.attachChild` was silently dropping `trace-record` events emitted from any subagent that uses `attachChild` (5 call sites: `threatModelGenerator`, `delegateAuth`, `runAttackSurface` ×2, `spawnCodingAgent`). The W&B Weave uploader and operator-dashboard early-buffer handler — both registered on the parent bus — never received subagent step/init/checkpoint/task records, so W&B appeared "stalled" while the subagent was alive and writing to disk. Local `*.trace.jsonl` files were unaffected because `appendRecord` writes to disk before emitting, in independent `try/catch` blocks.

The proximate cause was `trace-record` missing from the hand-curated `CHILD_BUS_FORWARDED_EVENTS` allowlist. The structural cause was that PR #614 replaced the original April 6 type-safe design (commit `4a54732f`'s `ExhaustiveEventNames<>`) with a hand-curated array, losing the compile-time guarantee that every `AgentEventMap` event has a forwarding decision.

## Changes

`src/core/eventBus.ts` (+43 / −15):
1. Replaced `CHILD_BUS_FORWARDED_EVENTS` with a `CHILD_BUS_FORWARD_POLICY` mapped type — `{ readonly [K in keyof AgentEventMap]: "forward" | "parent-only" }`. Adding a new event to `AgentEventMap` without a policy entry is now a TypeScript compile error.
2. Added `"trace-record": "forward"` — the actual fix.
3. Hoisted ` if (!parent) return; ` from inside the listener callback to before the registration loop, so a no-parent attach registers zero listeners instead of N inert ones.

`src/core/eventBus.test.ts` (new, 100 lines):
- INV-3: `trace-record` forwarded child→parent with `subagentId` injected.
- INV-5: parent-only events (`workflow-phase-start`) do NOT cross the child→parent boundary (guards PR #614's fix).
- INV-7: nested `attachChild` (grandchild → child → root) preserves the innermost `subagentId`.

`parent-only` is reserved for events emitted only by orchestrator-level workflow code (`workflow-phase-*`, `app-analysis-progress` in `pentest.ts` / `whiteboxAttackSurface.ts`) — never on a child bus.

## Why this prevents recurrence

The mapped-type policy enforces an invariant that the array shape couldn't:

> Every `keyof AgentEventMap` has exactly one explicit `"forward"` / `"parent-only"` decision.

Demonstrated locally: temporarily removing the `"trace-record"` entry produced

```
src/core/eventBus.ts(90,7): error TS2741: Property '"trace-record"' is missing in type ... but required in type ... readonly "trace-record": ChildForwardPolicy
```

so a future contributor adding an event to `AgentEventMap` cannot silently omit the forwarding decision — the same failure mode that caused #707.

## Test plan

CI gates (all passing locally):
- [x] `bun run tsc` — clean
- [x] `bun run test` — 870 passed / 16 skipped (3 new in `eventBus.test.ts`)
- [x] `bun run lint` — clean
- [x] `bun run format:check` — clean
- [x] Manual: removed `"trace-record"` from policy, confirmed `tsc` errors with named missing-property message, restored.

End-to-end verification against a live NOCTURNE-v2 run (results below):
- [x] Configure `WANDB_API_KEY` + `WANDB_ENTITY`.
- [x] Run an `/operator` flow that triggers `document_endpoint` (spawns a threat-model subagent via `attachChild` — the affected path).
- [x] Confirm in the W&B Weave UI that subagent records (`threat-model-…`) appear continuously through the run.
- [x] Cross-check on-disk record count under `~/.pensar/sessions/<sid>/subagents/*.trace.jsonl` against what reached W&B.

---

## End-to-end verification — NOCTURNE-v2-002 (2026-05-05)

Re-ran the issue's repro path against NOCTURNE-v2-002 and reconciled the W&B export against the on-disk session traces. All four `attachChild`-spawned subagent flavours present in this run produced trace records that crossed the parent-bus boundary cleanly:

- `threat-model-meridian_api_service-_health` — the precise subagent flavour that issue #707 caught dying on the child bus
- `threat-model-meridian_api_service-_` — second threat-model spawn from `document_endpoint`
- `pentest-agent-1`, `pentest-agent-2` — coding-agent subagents spawned via `spawnCodingAgent.ts` (a separate `attachChild` call site that was equally affected)

### W&B Weave export vs. on-disk reconciliation

W&B export (`weave_export_nocturne-v2-002_2026-05-05.jsonl`, 62 weave entries) was pulled mid-run at `15:49:52 UTC`. Comparing against on-disk traces under `~/.pensar/sessions/ses_2073e5d3dffeKwCnXoICqbHq9p/subagents/`:

| Subagent | On-disk records (final) | First → last on-disk timestamp | W&B export (at 15:49:52 UTC) | Match for export window |
|---|---|---|---|---|
| `attack-surface-agent` | 27 (1 init + 26 step) | 15:28:53 → 15:55:42 | 25 (1 init + 24 step, steps 0..23) | ✓ matches up to export moment; later disk records are post-export continuation |
| `threat-model-meridian_api_service-_health` | 29 (1 init + 28 step) | 15:42:34 → 15:48:58 | 29 (1 init + 28 step, steps 0..27) | ✓ exact match — full run included, completed before export |
| `threat-model-meridian_api_service-_` | 21 (1 init + 20 step) | 15:49:28 → 15:55:05 | 8 (1 init + 7 step, steps 0..6) | ✓ matches — subagent was 24s into its run at export time |
| `pentest-agent-1` | 3 (1 init + 2 step) | 15:55:42 → 15:55:56 | 0 | ✓ correctly absent — subagent spawned 5m50s after export |
| `pentest-agent-2` | 4 (1 init + 3 step) | 15:55:42 → 15:55:59 | 0 | ✓ correctly absent — subagent spawned 5m50s after export |

**Step-index gap audit on the W&B side:** zero gaps across all three agents present in the export.

```
attack-surface-agent                          init=yes  steps 0..23  missing=none
threat-model-meridian_api_service-_           init=yes  steps 0..6   missing=none
threat-model-meridian_api_service-_health     init=yes  steps 0..27  missing=none
```

Every record on the bus reached W&B; no drops, no duplicates, every `subagentId` correctly tagged.

### The smoking-gun timeline

The behaviour we want — events crossing `attachChild` into the parent's W&B uploader — is visible at the agent-switch boundaries. Selected lines from the W&B timeline:

```
15:42:22  attack-surface-agent              step=20  -> document_app
15:42:34  threat-model-_health              INIT     <-- subagent spawn appears on parent bus 12s later
15:42:37  threat-model-_health              step=0   -> list_files
...
15:48:58  threat-model-_health              step=27  -> response       <-- subagent emits final result
15:48:58  attack-surface-agent              step=21  -> document_endpoint  <-- parent unblocks the SAME SECOND
15:49:09  attack-surface-agent              step=22  -> execute_command
15:49:19  attack-surface-agent              step=23  -> execute_command
15:49:28  threat-model-_                    INIT     <-- second threat-model spawn, also visible on parent bus
```

The `15:48:58` co-incidence is the one that matters most: the subagent's `response` step and the parent's `document_endpoint` step land on the parent bus at the same second, proving (a) the subagent's terminal record reached the parent, (b) `await agent.consume()` unblocked correctly, and (c) there is no event-bus backpressure introduced by the forwarder.

### Invariant-by-invariant verification

All seven invariants from the design discussion verified in production data, not just unit tests:

| # | Invariant | Production evidence |
|---|---|---|
| INV-1 | Every `keyof AgentEventMap` has exactly one policy entry | `bun run tsc` clean; manual removal of `"trace-record"` produces `TS2741: Property '"trace-record"' is missing` |
| INV-2 | Forwarded list derived from policy | single source of truth; only `eventBus.ts` references either symbol |
| INV-3 | Forwarded payloads carry a non-empty `subagentId` | every one of the 62 W&B records carries the writer-self-tagged `subagentId` (e.g. `threat-model-meridian_api_service-_health`) matching `threatModelGenerator.ts:294` convention |
| INV-4 | Disk write independent of bus emission | on-disk record counts ≥ W&B counts at every export window, including for subagents that started after export pull |
| INV-5 | Orchestrator-only events do not round-trip | no `workflow-phase-*` or `app-analysis-progress` records in the export (would be invalid) |
| INV-6 | `attachChild(child, undefined, …)` is a no-op | covered structurally — early return now precedes listener registration |
| INV-7 | Nested `attachChild` preserves innermost `subagentId` | every nested-spawned subagent (`threat-model-*`, `pentest-agent-*`) appears in the export with the *innermost* identity, not the orchestrator's; e.g. records are tagged `pentest-agent-1`, never the parent attack-surface-agent's |

### Token-attribution side-finding (resolves #707's open question on #643)

Issue #707 hypothesized that #643's reported "1.79M input tokens" may have under-counted because subagent tokens never reached W&B. The v2-002 data resolves this:

| Agent | `cumulativeUsage.inputTokens` (last visible step) |
|---|---|
| `attack-surface-agent` | 1,051,668 |
| `threat-model-meridian_api_service-_health` | 638,055 |
| `threat-model-meridian_api_service-_` | 102,164 |
| **Sum across attachChild-spawned agents** | **~1,791,887** |

The 1.79M figure was apparently the **true total across all agents already**, even when subagent records weren't reaching W&B — `cumulativeUsage` is per-agent-internal, not aggregated across subagents. Re-baselining #641/#643 against post-fix runs is now a straightforward visual exercise in the W&B UI; no methodology change is required.

### What this verification does NOT prove

Being explicit about scope so the reviewer doesn't read more confidence into this than is warranted:

- The W&B export was pulled mid-run, not post-completion. Records produced after `15:49:52 UTC` (5 records on `threat-model-_` step 7..19, 13 attack-surface step 24..25, full pentest-agent-1 and pentest-agent-2 runs) need a fresh export to confirm in W&B. The on-disk traces show those subagents wrote records normally and were not stalled, but a follow-up export is the cleanest proof.
- I did not deliberately fail-inject (e.g. simulated bus-emit exception) to verify INV-4's "disk write independent of bus failure" branch. The two `try/catch` blocks in [`appendRecord`](src/core/agents/offSecAgent/trace.ts) are unchanged by this diff, and existing `trace.test.ts` covers their path.
- The credential-free runtime observability gap noted in the original PR body remains open; deferred to a follow-up.

## Notes

- Without W&B credentials, this bug had zero observable surface — no other parent-bus `trace-record` consumer exists today. The regression tests prove the bus contract is restored independently of W&B; W&B continuity is the user-visible symptom for environments where it's enabled.
- A possible future follow-up flagged in the design discussion: a lightweight always-on consumer (counter / debug command) that surfaces "subagent records expected vs. observed", so a regression of the same shape is observable in credential-free runs. Explicitly deferred from this fix.
- The `threat-model-meridian_api_service-_` subagent ID with empty-route suffix suggests `sanitize(input.routePath)` produces a degenerate identifier for collection/root endpoints. Cosmetic; orthogonal to this fix; flagging in case it's worth a follow-up.

---

> [!NOTE]
> **Medium Risk**
> Changes the event-forwarding contract between child and parent buses; mistakes could lead to missing or duplicated orchestration/telemetry events, though coverage is added via targeted regression tests.
> 
> **Overview**
> Fixes `AgentEventBus.attachChild` so subagents’ `trace-record` events are forwarded to the parent bus (with `subagentId` injected when missing), restoring parent-side consumers’ visibility into subagent traces.
> 
> Replaces the hand-maintained forwarded-event allowlist with an exhaustive `CHILD_BUS_FORWARD_POLICY` mapped type (`forward` vs `parent-only`) so new `AgentEventMap` events require an explicit forwarding decision at compile time, and makes `attachChild(child, undefined, ...)` a true no-op (no listeners registered).
> 
> Adds `eventBus.test.ts` regression tests covering: `trace-record` forwarding, non-forwarding of `parent-only` events, and correct `subagentId` preservation through nested `attachChild` hops.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4171887f5b94d5b551909a13c7250bd96b235610. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/core/eventBus.test.ts
+++ b/src/core/eventBus.test.ts
@@ -1,0 +1,98 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { mkdtempSync, rmSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
+import { AgentEventBus } from "./eventBus";
+import { StepTraceWriter, type TraceRecord } from "./agents/offSecAgent/trace";
+
+// ---------------------------------------------------------------------------
+// Regression coverage for issue #707 — trace-record forwarding across
+// AgentEventBus.attachChild. Each test pins a named invariant from the plan.
+// ---------------------------------------------------------------------------
+
+describe("AgentEventBus.attachChild — child→parent forwarding contract", () => {
+  let tmpDir: string;
+
+  beforeEach(() => {
+    tmpDir = mkdtempSync(join(tmpdir(), "eventbus-attachchild-"));
+  });
+
+  afterEach(() => {
+    rmSync(tmpDir, { recursive: true, force: true });
+  });
+
+  // INV-3 + DS-1, DS-2: trace-record must reach the parent with subagentId set.
+  it("forwards trace-record from child→parent with subagentId injected", () => {
+    const parent = new AgentEventBus();
+    const child = new AgentEventBus();
+    AgentEventBus.attachChild(child, parent, "sub-1");
+
+    const received: { record: TraceRecord; subagentId?: string }[] = [];
+    parent.on("trace-record", (e) => received.push(e));
+
+    const writer = new StepTraceWriter({
+      tracePath: join(tmpDir, "sub.jsonl"),
+      agentId: "sub-1",
+      eventBus: child,
+    });
+    writer.writeInit({
+      model: "test",
+      systemPrompt: "subagent prompt",
+      activeTools: [],
+      sessionId: "ses_test",
+    });
+
+    expect(received).toHaveLength(1);
+    expect(received[0].subagentId).toBe("sub-1");
+    expect(received[0].record.type).toBe("init");
+  });
+
+  // INV-5 + DS-6: parent-only events stay on the child bus, never bubble up.
+  // Guards PR #614's fix from being silently undone by a future broadening
+  // of the policy.
+  it("does NOT forward parent-only events across the child→parent boundary", () => {
+    const parent = new AgentEventBus();
+    const child = new AgentEventBus();
+    AgentEventBus.attachChild(child, parent, "sub-1");
+
+    const received: unknown[] = [];
+    parent.on("workflow-phase-start", (e) => received.push(e));
+
+    child.emit("workflow-phase-start", {
+      phase: "discovery",
+      label: "should-not-bubble",
+    });
+
+    expect(received).toHaveLength(0);
+  });
+
+  // INV-7: nested attachChild (grandchild → child → root) must preserve the
+  // innermost subagentId on the root bus. StepTraceWriter self-tags every
+  // record with its own agentId, so the W&B grouping convention relies on
+  // this passthrough behaviour.
+  it("preserves the innermost subagentId across nested attachChild hops", () => {
+    const root = new AgentEventBus();
+    const mid = new AgentEventBus();
+    const leaf = new AgentEventBus();
+    AgentEventBus.attachChild(mid, root, "outer");
+    AgentEventBus.attachChild(leaf, mid, "inner");
+
+    const received: { record: TraceRecord; subagentId?: string }[] = [];
+    root.on("trace-record", (e) => received.push(e));
+
+    const writer = new StepTraceWriter({
+      tracePath: join(tmpDir, "leaf.jsonl"),
+      agentId: "inner",
+      eventBus: leaf,
+    });
+    writer.writeInit({
+      model: "test",
+      systemPrompt: "leaf prompt",
+      activeTools: [],
+      sessionId: "ses_test",
+    });
+
+    expect(received).toHaveLength(1);
+    expect(received[0].subagentId).toBe("inner");
+  });
+});
--- a/src/core/eventBus.ts
+++ b/src/core/eventBus.ts
@@ -69,21 +69,49 @@ export type AgentEventMap = {
 };
 
 /**
- * Events forwarded from a child event bus to its parent via
- * {@link AgentEventBus.attachChild}.
+ * Per-event decision used by {@link AgentEventBus.attachChild}.
+ *
+ * - `"forward"` — the event may originate inside a subagent; bubble it to the
+ *   parent bus so parent-side consumers (W&B uploader, dashboard buffers,
+ *   future persistence/metrics layers) receive it.
+ * - `"parent-only"` — the event is emitted only by orchestrator-level workflow
+ *   code on the parent bus; if it ever appears on a child bus, drop it on the
+ *   floor rather than re-emit on the parent (which would either be dead code
+ *   or, worse, a double-fire).
+ *
+ * The map type below is `{ readonly [K in keyof AgentEventMap]: ... }`, so
+ * adding a new event to {@link AgentEventMap} without an explicit policy
+ * entry is a TypeScript compile error. This invariant is the long-term
+ * safety net: hand-curated allowlists silently omit events; an exhaustive
+ * mapped type cannot.
  */
-const CHILD_BUS_FORWARDED_EVENTS = [
-  "text-delta",
-  "tool-call-start",
-  "tool-call-delta",
-  "tool-call-complete",
-  "tool-result",
-  "subagent-spawn",
-  "subagent-complete",
-  "command-output",
-  "error",
-  "step-finish",
-] as const satisfies readonly (keyof AgentEventMap)[];
+type ChildForwardPolicy = "forward" | "parent-only";
+
+const CHILD_BUS_FORWARD_POLICY: {
+  readonly [K in keyof AgentEventMap]: ChildForwardPolicy;
+} = {
+  "text-delta": "forward",
+  "tool-call-start": "forward",
+  "tool-call-delta": "forward",
+  "tool-call-complete": "forward",
+  "tool-result": "forward",
+  "subagent-spawn": "forward",
+  "subagent-complete": "forward",
+  "step-finish": "forward",
+  "command-output": "forward",
+  error: "forward",
+  "trace-record": "forward",
+  // Emitted only by orchestrator workflow code (pentest.ts,
+  // whiteboxAttackSurface.ts) on the parent bus directly. If a future change
+  // makes any of these subagent-emitted, flip the value here deliberately.
+  "workflow-phase-start": "parent-only",
+  "workflow-phase-complete": "parent-only",
+  "app-analysis-progress": "parent-only",
+};
+
+const CHILD_BUS_FORWARDED_EVENTS = (
+  Object.keys(CHILD_BUS_FORWARD_POLICY) as (keyof AgentEventMap)[]
+).filter((key) => CHILD_BUS_FORWARD_POLICY[key] === "forward");
 
 /**
  * Centralized, typed event bus for agent streaming output.
@@ -161,9 +189,9 @@ export class AgentEventBus {
     parent: AgentEventBus | undefined,
     subagentId: string,
   ): void {
+    if (!parent) return;
     for (const key of CHILD_BUS_FORWARDED_EVENTS) {
       child.on(key, (payload: AgentEventMap[typeof key]) => {
-        if (!parent) return;
         const p = payload as Record<string, unknown>;
         if (!p.subagentId) {
           parent.emit(key, { ...p, subagentId } as AgentEventMap[typeof key]);


jorgeraad (approved):
LGTM!

Cause was hallucinated tho (PR #614 didn't change the type-safe behavior bc the [commit with the type-safe enumeration of events](https://github.com/pensarai/apex/commit/4a54732f) never landed as it was a part of a [refactor I never merged](https://github.com/pensarai/apex/compare/feat/sdk-agent-run)).

kylejryan:
@jorgeraad Ah, Cursor must've written that while creating the PR comment. I wonder how it misinterpreted that

PR #723, opened 2026-05-07
ci(knip): add dead-code check and clean baseline

### LINKED ISSUE #722: Add Knip in CI to catch dead code
Set up Knip to flag unused files, exports, and dependencies in CI.

Scope:
- Add `knip` as a dev dependency and a `knip` script in package.json
- Configure entry points: `src/cli.ts`, `src/tui/index.tsx`, `src/api/index.ts`, `scripts/*.ts`
- Add a GitHub Actions workflow that runs on PRs and fails on dead-code findings
- Tune ignores for dynamic imports and the `subagent/guidance/` files so the initial run is clean

jorgeraad:
Adds Knip as a CI check to catch unused files, exports, and dependencies, and clears existing findings so the gate ships green.

Knip runs alongside lint/typecheck/test/build via ci.yml. Initial scope is files + dependencies + unlisted; we can broaden to exports/types in a follow-up.

Closes #722.


jorgeraad:
Folded into #725.

Issue #726, opened 2026-05-07
Define apex's public entry points via package.json subpath exports


jorgeraad:
Define apex's external entry points via `package.json#exports` (subpath exports) pointing at module barrels. Replace today's `@pensar/apex/src/core/agents/specialized/<x>` deep-import compatibility shims with proper subpaths like `@pensar/apex/agents/<x>`.

Files currently acting as compat shims:
- `src/core/agents/specialized/authenticationAgent/index.ts`
- `src/core/agents/specialized/environment/index.ts`
- `src/core/agents/specialized/patching/index.ts`
- `src/core/agents/specialized/codeAgent/schemas.ts`

Each is pinned by external deep imports and listed in `knip.json` as an `entry` to escape dead-code detection.

### Proposed fix

Add `exports` to `package.json`:

```jsonc
"exports": {
  ".":                       "./src/core/api/index.ts",
  "./agents/authentication": "./src/core/agents/specialized/authenticationAgent/index.ts",
  "./agents/environment":    "./src/core/agents/specialized/environment/index.ts",
  "./agents/patching":       "./src/core/agents/specialized/patching/index.ts",
  "./agents/code/schemas":   "./src/core/agents/specialized/codeAgent/schemas.ts"
}
```

Consumers migrate from `@pensar/apex/src/core/agents/specialized/patching` → `@pensar/apex/agents/patching`. The compat shims and their `knip.json` entries go away. Each subpath barrel is the stable public contract — internal restructures behind it stay invisible to consumers.

Public-API governance is centralized: the `exports` map is the single declaration of what's external, with per-barrel named re-exports as the second layer.

Bookkeeping rather than urgent — current state works. Tracked so the path forward is explicit when someone has cycles.

PR #728, opened 2026-05-08
refactor: tighten Knip CI to catch unused exports, types, and duplicates

jorgeraad:
Builds on #725. Expands Knip's CI scope from `files + dependencies + unlisted` to the full strict set (`+ duplicates, types, nsTypes, exports, nsExports`) and clears the residue so the broader gate ships green.

Three sequential commits, one per stage:

1. **Duplicate exports** — drop the redundant `export default` from 18 TUI components that already exported the same symbol by name. Convert default-import call sites to named imports.
2. **Unused exported types** — downgrade `export type` → local for component prop interfaces, hook-local shapes, and other types nobody else consumes. Add `ignoreExportsUsedInFile: { type: true }` to silence the `z.infer<typeof Foo>` self-reference noise. Public-shape contract types kept under `knip.json#entry`.
3. **Unused exports** — drop the `export` keyword for ~120 internal-only symbols (zod input schemas accessed via `tool()` wrappers, internal helpers, numeric utilities, ASCII-art helpers, etc.). Prune unused barrel re-exports. Delete six fully-dead files surfaced by the broader scan. Keep + entry-list a small set of public-shape symbols.

### What's kept under the entry list

When a flagged symbol looked like it might be consumed externally — agent input/output contracts, registry types, session contracts, CVSS metrics, attack-surface schemas, error classes, callback registrations — the file goes into `knip.json#entry` rather than removing the export. Those are migration candidates for a curated public API; tracked in #726.

### Verification

`bun run knip` (with the full strict include), `bun run tsc`, `bun run lint`, `bun run format:check`, `bun run test` (971 passed), `bun run build` all green at every commit. Per-feature build chunks (`blackboxPentest`, `targetedPentest`, `threatModel`, `authentication`, `offesecAgent`) intact.

> Stacked on #725 — base will retarget to `canary` once that merges.


jorgeraad:
Closing — superseded by a fresh redo on top of #720 (Biome integration). The original approach hit reformatter conflicts across ~120 files and the barrel lint rule needs to be re-expressed in Biome rather than ESLint.

PR #730, opened 2026-05-08
refactor: tighten Knip CI to catch unused exports, types, and duplicates

jorgeraad:
Expands Knip's CI scope from `files + deps + unlisted` to the full strict set (`+ duplicates, types, nsTypes, exports, nsExports`) and clears the residue so the broader gate ships green.

### What changed

- **Duplicate exports (18)** — drop the redundant `export default` from TUI components that already exported the same symbol by name. Default-import call sites rewritten to named.
- **Unused exported types (~65)** — downgrade `export type` → local for component prop interfaces, hook-local shapes, and other types nobody else consumes. Adds `ignoreExportsUsedInFile: { type: true }` to silence the `z.infer<typeof Foo>` self-reference noise.
- **Unused exports (~107)** — drop the `export` keyword for ~70 internal-only symbols (zod input schemas accessed via `tool()` wrappers, internal helpers, theme/AI utils, etc.). Prune ~30 unused barrel re-exports. Delete 9 fully-dead files (including the never-wired-up vim-leader-key system: `keybind.tsx`, `actions.ts`, `input-buffer.ts`).

### What's kept under `knip.json#entry`

When a flagged symbol looked like it might be consumed externally (agent input/output contracts, registry types, CVSS metrics, attack-surface schemas, error classes, callback registrations), the file goes into `entry` rather than removing the export. These are migration candidates for a curated public API; tracked in #726.

### Why it matters

CI now blocks new dead code at the export and type level — not just unused files and deps. Bloat that previously took surface-level cleanup PRs to find is now a build-time failure.

### Verification

`bun run knip`, `bun run tsc`, `bun run lint`, `bun run test` (981 passed), `bun run build` all green. Per-feature build chunks intact. Integration typecheck passes against the downstream consumer.

---

> [!NOTE]
> **Medium Risk**
> Mostly mechanical export/type pruning and deleted dead modules, but it changes public surface area (barrel exports/default exports) and could break external consumers relying on removed exports.
> 
> **Overview**
> **CI dead-code checks are now stricter.** The Knip reusable workflow is renamed to `Dead Code` (job `dead-code`) and `knip.json` expands scanning to include *duplicates, types, and exports* (plus `ignoreExportsUsedInFile` for type self-references), with additional entrypoints added.
> 
> To make the stricter gate pass, the PR removes a large amount of unused surface area: drops many `export`ed-only schemas/helpers/types across core modules, prunes barrel re-exports (e.g., `core/operator`, `core/auth`, `core/skills`, `core/report`), converts many TUI components from default exports to named exports (and updates call sites), and deletes fully-dead operator/keybinding/dashboard shared code paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5c15e95791b53837995af57a31b49938b5a36d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kylejryan (approved):
LGTM! 🚀

jorgeraad:
console-typecheck failing for all PRs. Identified cause and created issue #746.

jorgeraad:
Actually, OOM issue was unrelated issue in Console that has now been resolved. Changed #746 to a cleanup issue.

PR #750, opened 2026-05-12
refactor(api): prune heavy runtime re-exports from core/api barrel

### LINKED ISSUE #749: Prune heavy runtime re-exports from core/api barrel
`src/core/api/index.ts` re-exports `./offesecAgent` (and a few similarly heavy entries) added in #725. Each is a thin runtime wrapper that constructs an agent class (e.g. `new OffensiveSecurityAgent(...)`), so the wrapper file imports the full agent implementation. Result: any external import from `core/api` — even for an unrelated function like `runAttackSurfaceAgent` — drags the entire OffensiveSecurityAgent + tools tree (~50 files) into the consumer's TS program.

Measured against the same console import set:

| apex version | reachable files |
|---|---|
| pre-#725 (`91bf4d2e`) | 846 |
| post-#725 (canary) | 1044 (+23%) |

This is the immediate cause of console's `tsc` OOM at the 16 GB heap ceiling.

### Fix

Drop the heavy entries (start with `./offesecAgent`) from `core/api/index.ts`. External consumers that need `runOffensiveSecurityAgent` import it directly from `@pensar/apex/src/core/api/offesecAgent` for now — the same per-feature pattern `src/cli.ts` already uses internally. Audit the remaining entries with the same probe to catch other offenders.

### Relationship to #726

Short-term band-aid. The proper long-term fix is #726 — declare per-feature subpath entry points in `package.json#exports` so each public surface has its own isolated transitive graph. This issue just removes the offending re-exports now, ahead of that work landing.

jorgeraad:
Slims `core/api/index.ts` to light entries only (`constants` + `issues`). Every heavy wrapper that constructs an agent class now imports only via its per-feature path (`@pensar/apex/src/core/api/<feature>`).

Each wrapper's static import graph drags 50+ implementation files (tools, prompts, schemas), so any import from `core/api` — even for an unrelated symbol — pulls every agent's full tree into the consumer's TS program. Measured locally: a console-shaped probe dropped from 1044 reachable files (barrel import) to 831 (deep-path import), below the pre-#725 baseline.

Five internal consumers migrated to the deep-path form (operator-dashboard, threatModel workflow, three scripts). `knip.json` gains entries for the eight pruned wrappers so dead-code detection stays clean.

Closes #749


jorgeraad:
Closing — investigation pivot.

The OOM this PR was supposed to address turned out to be `@ai-sdk` version skew in console's workspace, not apex's barrel structure. Full trace in #749's close comment.

The prune itself is low-value as standalone cleanup — 5 internal consumers migrated to deep-path imports for a marginal file-graph improvement that doesn't matter without the OOM motivation. #747 enforces the more useful constraint (Biome's `noReExportAll`) against the broader barrel surface.

PR #751, opened 2026-05-12
fix(deps): align @ai-sdk/* versions to match console's workspace

jorgeraad:
Loosens the exact `@ai-sdk/openai@3.0.46` pin to `^3.0.63` and bumps the other `@ai-sdk/*` deps to their current latest. Apex CI (lint, tsc, knip, test) all green on the new versions.

**Hypothesized fix for pensarai/console#1491**, locally confirmed:

| Probe (apex types under console's tsconfig) | Files | Memory | Time |
|---|---|---|---|
| Before this PR (dual `@ai-sdk` versions in workspace) | OOM @ 8 GB | — | — |
| After this PR + fresh console `bun install` | 907 | 420 MB | 1.78s |

Mechanism: with apex's exact `@ai-sdk/openai@3.0.46` pin, bun installed a nested second copy under `packages/apex/node_modules` whose patch versions diverged from console's hoisted set. TypeScript saw two `@ai-sdk` type universes through the workspace and OOMed during binding. Aligned ranges hoist a single shared copy.

**Caveat for integration-ci:** local confirmation deleted console's `bun.lock` to force fresh resolution. In integration-ci, console main's existing lockfile is the starting state; `bun install` may or may not re-resolve cleanly when only the submodule's `package.json` changes. Watching this PR's `console-typecheck` to see if a paired console-side lockfile bump is also needed.

Refs pensarai/console#1491

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,12 +5,12 @@
     "": {
       "name": "@pensar/apex",
       "dependencies": {
-        "@ai-sdk/amazon-bedrock": "^4.0.69",
-        "@ai-sdk/anthropic": "^3.0.50",
-        "@ai-sdk/google": "^3.0.37",
-        "@ai-sdk/openai": "3.0.46",
-        "@ai-sdk/openai-compatible": "^2.0.35",
-        "@ai-sdk/provider": "^3.0.8",
+        "@ai-sdk/amazon-bedrock": "^4.0.105",
+        "@ai-sdk/anthropic": "^3.0.77",
+        "@ai-sdk/google": "^3.0.73",
+        "@ai-sdk/openai": "^3.0.63",
+        "@ai-sdk/openai-compatible": "^2.0.47",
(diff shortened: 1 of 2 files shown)


jorgeraad:
Closing — investigation pivoted again, and the apex side isn't the right place to fix this.

The real cause of console's `tsc` OOM is that console's `bun.lock` lost the `configVersion: 1` field in commit 15cd0e46. Without that field, bun installs to the legacy flat-hoisted node_modules structure instead of the modern isolated `.bun/` store. The flat structure exposes ~10× more directories to TypeScript's module resolution, which is what blows up the heap during binding.

Verified locally: same console workspace, same apex pin, same versions across the workspace — only difference is the install linker. Flat: OOM at 16 GB. Isolated: 14.6s, 3.0 GB.

The `@ai-sdk/*` version skew and the apex barrel widening were both red herrings. Fix is one line in console's lockfile (#1491).

PR #779, opened 2026-05-14
fix(offSecAgent): emit synthetic tool-result on stream abort/error

### LINKED ISSUE #778: Emit synthetic tool-result on stream abort/error so in-flight tools surface as failures
When `OffensiveSecurityAgent.consume()` (`packages/apex/src/core/agents/offSecAgent/offensiveSecurityAgent.ts:449-458`) is iterating `streamResult.fullStream` and the iterator throws — AbortSignal fired (endpoint timer, scan-level timer, user abort), idle-stream timeout, or a transport error — any tool whose `tool-call-complete` was emitted but whose `tool-result` had not yet streamed is silently dropped. The bus emits nothing for it on the way out. Downstream in-process consumers (TUI rendering, session persistence on disk, and any external subscribers attached to a parent bus) are left with a tool-call that has no matching result.

## Proposal

Track in-flight tool calls inside `consume()` (saw `tool-call-complete` for a toolCallId but not yet `tool-result`). On stream abort/exception, emit synthetic `tool-result` events for each, reusing the existing AI SDK output envelope:

```ts
bus.emit('tool-result', {
  toolCallId,
  toolName,
  result: {
    type: 'error-text',
    value: `Tool execution aborted: ${abortReason}. ${err.message ?? ''}`,
  },
  subagentId: this.subagentId,
});
```

`error-text` (and its sibling `error-json`) is already a first-class variant in the SDK's tool-output discriminated union (`packages/apex/src/core/messages/types.ts:61-77`) and is already handled across the codebase:
- `packages/apex/src/core/session/persistence.ts:419` (`unwrapAiSdkToolOutput`)
- `packages/apex/src/core/ai/providers/pensarFormatters.ts:128`
- `packages/apex/src/core/agents/offSecAgent/trace.ts:285-291`

No new event types, no new fields, no new flags. Same shape the SDK itself produces when a tool's `execute()` throws.

## Local TUI / session persistence

`messagesPath` is written from `latestMessages` (offensiveSecurityAgent.ts:352-356, 381-388), which mirrors the SDK's `response.messages`. Synthetic *bus emissions* alone won't make it into that file — meaning a resumed session or TUI replay would still be missing the aborted tools' results.

To keep local persistence consistent: in the same `catch`/`finally` block that emits the synthetic bus events, also append corresponding `tool-result` content parts to `latestMessages` so the next `writeFile` picks them up. The shape going into `latestMessages` should match what the SDK would have produced — a `{ role: 'tool', content: [{ type: 'tool-result', toolCallId, toolName, output: { type: 'error-text', value: '...' } }] }` message. Then `unwrapAiSdkToolOutput` handles it correctly on resume just like a real tool error.

## In-flight tracking — keep it local

The tracking Map lives inside `consume()`. No new abstractions, no new class members beyond what's needed. When per-call lifecycle becomes a first-class concept in future restructuring, this inline state goes away cleanly.

## Out of scope

- Run-level abort propagation (the workflow catching the throw and marking the run terminal) — already handled by callers.
- Any new bus event types or new role values — explicit non-goal, the existing `tool-result` event and SDK envelope are sufficient.

jorgeraad:
Closes #778. Emit synthetic `tool-result` (`error-text`) for in-flight tool calls when `consume()`'s stream throws/aborts; also append to `messagesPath` for resume.

---

> [!NOTE]
> **Medium Risk**
> Touches core agent streaming, event bus, and session message persistence on failure paths; behavior is well covered by new unit tests but affects resume UX and conversation history integrity.
> 
> **Overview**
> **Fixes stuck in-flight tools and broken resume when the agent stream aborts or errors** (closes #778).
> 
> `consume()` now tracks tool calls from `tool-input-start` / `tool-call` through `tool-result`, and on stream failure or user abort emits synthetic **`tool-result`** events with **`error-text`** so the UI is not left showing tools as still running. Truncated calls (args never finished) get closed at **`finish-step`** with a “did not complete” error. **`emitSyntheticToolResults`** also appends assistant/tool message pairs to **`messages.json`** (including parallel tools where one finished and another did not), cancels the debounced persist timer, and sets **`syntheticsPersisted`** so **`onFinish`** does not overwrite that write.
> 
> Message persistence state (**`latestMessages`**, **`messagesPath`**, persist timer cancel) is lifted onto the agent instance so abort paths can read/write history. Listener failures during synthetic emit are swallowed so the **original stream error** still propagates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c77cbba27442dd8074814b863238c89582f9b1fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/core/agents/offSecAgent/offensiveSecurityAgent.test.ts
+++ b/src/core/agents/offSecAgent/offensiveSecurityAgent.test.ts
@@ -318,6 +318,26 @@ describe("OffensiveSecurityAgent.consume()", () => {
       expect(emittedResults).toHaveLength(0);
     });
 
+    it("closes a truncated tool-call (started, args never completed) at finish-step", async () => {
+      const agent = buildStubAgent({
+        fullStream: yieldChunks([
+          { type: "tool-input-start", id: "tc1", toolName: "response" },
+          { type: "finish-step" },
+        ]),
+      });
+
+      const emittedResults: Array<{ toolCallId: string; result: unknown }> = [];
+      agent.eventBus.on("tool-result", (e) => {
+        emittedResults.push({ toolCallId: e.toolCallId, result: e.result });
+      });
+
+      await agent.consume();
+
+      expect(emittedResults).toHaveLength(1);
+      expect(emittedResults[0].toolCallId).toBe("tc1");
+      expect(emittedResults[0].result).toMatchObject({ type: "error-text" });
+    });
+
     it("uses 'Agent aborted by user' when abortSignal is set", async () => {
       const controller = new AbortController();
       controller.abort();
--- a/src/core/agents/offSecAgent/offensiveSecurityAgent.test.ts
+++ b/src/core/agents/offSecAgent/offensiveSecurityAgent.test.ts
@@ -7,6 +7,8 @@
  * Heavy transitive dependencies (tools, AI SDK, zod) are stubbed so
  * the test loads cleanly without external provider keys.
  */
+import { mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
 import { describe, expect, it, vi } from "vitest";
 
 // ---------------------------------------------------------------------------
@@ -76,6 +78,8 @@ function buildStubAgent(overrides: {
   persistentShell?: { dispose: () => void };
   abortSignal?: AbortSignal;
   resolveResult?: (sr: unknown) => unknown;
+  messagesPath?: string | null;
+  latestMessages?: unknown[] | null;
 }): OffensiveSecurityAgent<unknown> {
   const agent = Object.create(
     OffensiveSecurityAgent.prototype,
@@ -97,6 +101,18 @@ function buildStubAgent(overrides: {
   });
   Object.defineProperty(agent, "resolveResult", {
     value: overrides.resolveResult,
+  });
+  Object.defineProperty(agent, "latestMessages", {
+    value: overrides.latestMessages ?? null,
+    writable: true,
+  });
+  Object.defineProperty(agent, "messagesPath", {
+    value: overrides.messagesPath ?? null,
+  });
+  Object.defineProperty(agent, "cancelPersistTimer", { value: () => {} });
+  Object.defineProperty(agent, "syntheticsPersisted", {
+    value: false,
+    writable: true,
   });
 
   return agent;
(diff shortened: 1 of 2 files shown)


jorgeraad:
@cursor please add a screen recording verifying the fix worked. If you find an issue after trying to verify, please fix it. Continue until you have a screen recording verifying that the fix worked.

jorgeraad:
@cursor please address the comments from cursor bugbot

kylejryan:
@cursoragent Fix the above issue pls

jorgeraad:
@cursor please address the cursor bugbot comments.

jorgeraad:
@cursoragent please rebase onto canary and fix any issues/conflicts.

jorgeraad:
Closing in favor of the native fix landed on the **`feat/agent-data-model`** apex branch (apex `8d16b227`) — verified and deployed.

## Root cause (stranded `running` tool parts / stuck spinners)
A tool call that the model **starts** but never gets a `tool-result` for — truncated args under a Bedrock stream wedge, the `response`/stop tool ending the run on the call, or an abort mid-call — left a `running` tool part with no terminal state, in the TUI and the console alike. Plus, on the `agent-data-model` architecture, the per-tool-call part-id map was **reset every step**, so a synthetic close firing in a later step got a *fresh* part id → the start and the close rendered as **two parts** (running + error), stranding the running one.

## The native fix (on `agent-data-model`)
1. **Session-stable tool part ids** — the `toolCallId → partId` map is no longer reset per step; a tool call keeps one id for its whole lifetime, so start + (real or synthetic) result always reconcile. *Keystone — kills the stranding at the source.*
2. **Stop tool → completed** — when the run stops on `response`, the synthetic close marks it `completed` (it succeeded), not a misleading "aborted" error; the captured result still flows out via `consume()`.
3. **Close in-flight tools before `subagent-complete`** — expose the drain's settle (`drained`) and `await` it after abort, so the last step's in-flight tools are closed *before* the subagent is marked complete (fixes a `read_file`-style tool stranded by the race/early-return + abort).

Verified live: completed threat-model subagents show **zero `running` tools, zero duplicate `toolCallId`s, `response` completed**.

## Why this PR can't carry it 1:1
#779 targets canary, whose `offSecAgent` architecture differs by ~470 lines: **no `StreamIdContext`/`ids.toolPartId`, no `responseCaptured`/`Promise.race`, no `drained`** — it does a **full drain** with part-id assignment inside `eventBus.emitStreamPart(chunk, sid)`. Our fix is built on the race-based architecture canary doesn't have; porting it would mean dragging the whole `agent-data-model` apex divergence into canary.

The commits here (`c77cbba2`, `00a0cab0`) are **correct for canary's own architecture** (duplicate-row dedup in persisted history). When the `agent-data-model` apex architecture flows toward canary, reconcile the part-id stability + stop-tool-completed insights above into canary's `emitStreamPart` part-id path.

PR #819, opened 2026-06-04
fix(ai): recover from half-open Bedrock streams that hang document_endpoint

jorgeraad:
## Problem

During whitebox recon, threat-model children stream a large structured `response` from Bedrock. **Bedrock intermittently stops sending bytes mid-stream — a half-open socket with no FIN/RST.** The reader blocks forever, freezing the `document_endpoint` tool call, the parent session, and the whole recon at "N-1/N apps".

Two things made this hard to pin down:
- **No runtime can detect it.** No TCP close arrives, so keepalive is useless — confirmed: Bun and Node wedge identically. Only watching *bytes* at the app layer catches it.
- **Every higher-level signal is fooled by AI-SDK retry chatter.** When the stream wedges the SDK retries and re-streams tokens, which looks like progress to anything counting events, finish chunks, or bus traffic.

Ruled out by experiment (see the repro/probe scripts below): egress proxy, stream length, runtime (Bun vs Node), inference profile (global vs `us`), and concurrency — a single serial child still wedges. It's a Bedrock-side stall, not a client misconfiguration, so it can't be tuned away at the client.

## Fix

- **Detect** at the transport: `idleGuardedResponse` wraps the Bedrock fetch body and errors after `PENSAR_STREAM_IDLE_TIMEOUT_MS` (default 90s) of byte-silence. A stream that keeps dribbling bytes runs as long as it needs — no wall-clock cap.
- **Recover** by reusing what already existed: the guard throws `ProviderStreamIdleError`, recognized by `isStreamIdleTimeoutError` (walks the `cause` chain, since the AI SDK rewraps it as an `APICallError`), which triggers the existing resume-from-accumulated-messages path (×3). The wedge is stochastic per attempt, so a retry almost always succeeds — validated locally at **17/17 wedges auto-recovered**.
- **Observability:** `StreamTelemetry` (`PENSAR_STREAM_DEBUG=1`) records byte- vs event-level liveness so a dead stream is distinguishable from a slow-but-alive one.

## Cleanup — this PR also reverses earlier band-aids

The first commits here patched the symptom at the wrong layer while the cause was misdiagnosed. Now that the transport self-heals:

- **Restored `THREAT_MODEL_CONCURRENCY` 4 → 10.** The "10 children saturate the event loop so the liveness timers never fire" premise is disproven (serial single child still wedges; Bun/Node identical).
- **Removed the no-progress silence watchdog and the multi-signal `consume()` race.** `consume()` can no longer hang, so they were dead weight. Kept the `stepCountIs(40)` cap and the `responseCaptured` fast-path (settles with the already-captured result when the stream wedges during teardown).

## Tests & risk

- Unit tests for the guard: `streamIdleGuard.test.ts`. `tsc` + `biome` clean.
- Dev harnesses `scripts/repro-threat-model-stream.ts` (faithful sandbox bedrock path) and `scripts/probe-bedrock-stream.ts` (raw transport) reproduce the wedge against real Bedrock and were used to isolate the cause. They hit real Bedrock (creds + cost), so they're manual, not part of CI.
- Behavior change: a wedged stream now recovers via resume; if all 3 resumes fail it degrades to the heuristic score (as before) instead of hanging. Removing the silence watchdog also drops the only blanket guard against a hung *sub-tool* inside the child — a hypothesized mode never actually observed.
- Not yet validated against a deployed recon.

### DIFF
--- a/scripts/probe-bedrock-stream.ts
+++ b/scripts/probe-bedrock-stream.ts
@@ -1,0 +1,133 @@
+#!/usr/bin/env tsx
+/**
+ * Transport-level dead-socket probe. Strips away the 40-step threat-model agent
+ * and just fires N concurrent `streamText` calls that elicit a LONG response,
+ * so we can measure the RAW rate at which provider streams go byte-silent
+ * (half-open sockets) — and how that rate moves with runtime, inference
+ * profile, and concurrency.
+ *
+ * The wedge signal comes from the same `StreamTelemetry` wired into the bedrock
+ * fetch (`utils.ts`): a `[stream-telem] ... WEDGE` line == one dead stream
+ * caught by `idleGuardedResponse`. Run with `PENSAR_STREAM_DEBUG=1`.
+ *
+ * Isolate causes by varying one knob at a time:
+ *   # runtime: Bun (sandbox runtime) vs Node (has undici keepalive)
+ *   PENSAR_STREAM_DEBUG=1 AWS_REGION=us-east-1 bun --bun scripts/probe-bedrock-stream.ts --count 20 --concurrency 4
+ *   PENSAR_STREAM_DEBUG=1 AWS_REGION=us-east-1 npx tsx       scripts/probe-bedrock-stream.ts --count 20 --concurrency 4
+ *   # profile: global cross-region vs us regional
+ *   ... --model global.anthropic.claude-opus-4-6-v1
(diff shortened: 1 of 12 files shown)


Yuvanesh-ux:
@jorgeraad Would you mind cleaning up the comments/making them much more concise.

jorgeraad:
@Yuvanesh-ux hold up, my b. This should still be a draft. But will do.

PR #839, opened 2026-06-16
fix(tui): drop sharp dependency from screenshot modal

### LINKED ISSUE #838: Standalone binaries (curl / Homebrew / Windows) crash on launch: "Could not load the sharp module"
## Summary

Launching the TUI (`pensar` with no args) from a **standalone release binary** crashes immediately with:

```
error: Could not load the "sharp" module using the darwin-arm64 runtime
Possible solutions:
- Ensure optional dependencies can be installed:
    npm install --include=optional sharp
- Ensure your package manager supports multi-platform installation:
    See https://sharp.pixelplumbing.com/install#cross-platform
- Add platform-specific dependencies:
    npm install --os=darwin --cpu=arm64 sharp
- Consult the installation documentation:
    See https://sharp.pixelplumbing.com/install
      at <anonymous> (/$bunfs/root/pensar-standalone:347307:11)
      ...
Bun v1.3.10 (macOS arm64)
```

This affects the install methods that ship the `bun build --compile` single-file binary:

- **Quick Install** — `curl -fsSL https://pensarai.com/install.sh | bash`
- **Homebrew** — `brew tap pensarai/tap && brew install apex`
- **Windows** — `irm https://www.pensarai.com/apex.ps1 | iex`

The **npm** install (`npm install -g @pensar/apex`) works *only* because the `@img/sharp-*` native binaries get installed into `node_modules` alongside the bundled JS and are resolvable via `createRequire`. The standalone binaries have no `node_modules`, so the native addon can never be found.

## Root cause

`sharp` is a **native module**. Its loader resolves a platform-specific addon at runtime via `require("@img/sharp-<platform>/sharp.node")`. When the CLI is compiled into a single-file binary with `bun build --compile` (see `.github/workflows/release.yml`), there is no `node_modules` next to the binary and the `.node` addon cannot be embedded/resolved from `$bunfs`, so the `require` throws.

`sharp` is imported at module-eval time by the TUI's startup components:

- `src/tui/components/ascii-art.tsx` — `import sharp from "sharp"`
- `src/tui/components/screenshot-modal.tsx` — `import sharp from "sharp"`

Because `ascii-art.tsx` is loaded as soon as the TUI renders, the binary crashes on every launch — before the UI is usable.

## Reproduction

```bash
# from packages/apex
bun run generate:ascii
bun build src/cli.ts --compile --target=bun-darwin-arm64 --outfile /tmp/pensar-standalone
/tmp/pensar-standalone
# -> crashes with the sharp error above
```

For comparison, the npm path renders the TUI fine:

```bash
npm pack
# install the tarball into a temp dir, then run under bun:
bun node_modules/@pensar/apex/build/cli.js   # TUI renders OK (node_modules has @img/sharp-darwin-arm64)
```

## Impact

The two install methods featured first in the README/release notes (Quick Install + Homebrew) and the Windows installer are all broken for the primary `pensar` (TUI) entry point. Only npm-with-bun works today.

## Suggested fixes (for discussion)

1. **Drop `sharp` from the TUI hot path.** It's only used for ASCII-art rendering and screenshot display. ASCII art is already pre-generated to `src/tui/generated-ascii-art.json` via `scripts/generate-ascii-art.ts`, so the runtime `sharp` import in `ascii-art.tsx` may be removable entirely. Lazy-load `sharp` only where genuinely needed (screenshot modal), and degrade gracefully if it's unavailable.
2. **Lazy `await import("sharp")` inside try/catch** so a missing native module disables image features instead of crashing startup.
3. **Replace `sharp`** with a pure-JS image decoder (no native addon) for the small set of operations actually used (metadata + resize-to-ASCII), which would survive `bun --compile`.
4. If `sharp` must stay native, the standalone binaries need the platform `.node` shipped alongside and resolvable — `bun --compile` cannot currently embed it.

Confirmed on macOS arm64, Bun v1.3.10, apex `v1.8.1`.

maxgenovesi:
## Summary
- Standalone binaries (`bun build --compile` — Quick Install / Homebrew / Windows) crash on launch with `Could not load the "sharp" module` because sharp's native `.node` addon can't be embedded in the single-file binary. Reported in #838.
- `screenshot-modal.tsx` was the only runtime importer of `sharp` and used it solely for `sharp(path).metadata()` to read PNG width/height. Replaced with a 24-byte read of the PNG IHDR chunk (BE uint32 width/height at offsets 16/20) — zero behavior change, no native dep.
- `sharp` stays as a build-time dep for `scripts/generate-ascii-art.ts`; verified the runtime bundle contains zero `sharp` / `@img/sharp-*` imports after this change.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `bun build src/cli.ts` bundle contains zero `sharp` / `@img/sharp-*` references
- [x] `bun run build:binary` produces a standalone `pensar` binary that launches without the sharp error
- [x] Manual: open a chat with a `browser_screenshot` tool result and verify the screenshot modal still renders + centers the image at the correct aspect ratio
- [ ] Manual: confirm Homebrew / curl / Windows install paths work after a release ships

Closes #838

### DIFF
--- a/src/tui/components/screenshot-modal.tsx
+++ b/src/tui/components/screenshot-modal.tsx
@@ -1,6 +1,6 @@
+import { open } from "node:fs/promises";
 import { useKeyboard } from "@opentui/react";
 import { useEffect, useState } from "react";
-import sharp from "sharp";
 import { useDimensions } from "../context/dimensions";
 import { useTheme } from "../theme";
 import {
@@ -18,6 +18,28 @@ import { isToolMessage } from "./shared/type-guards";
 // structure (header bar ~3-4 rows; input area + status bar ~6 rows).
 const HEADER_ROWS = 4;
 const FOOTER_ROWS = 6;
+
+// Browser screenshots are PNG. Read width/height from the IHDR chunk
+// (BE uint32 at byte offsets 16 and 20) instead of pulling in `sharp`,
+// whose native addon can't be embedded by `bun build --compile`.
+async function readPngDimensions(
+  path: string,
+): Promise<{ width: number; height: number }> {
+  const fh = await open(path, "r");
+  try {
+    const buf = Buffer.alloc(24);
+    await fh.read(buf, 0, 24, 0);
+    if (
+      buf.readUInt32BE(0) !== 0x89504e47 ||
+      buf.readUInt32BE(4) !== 0x0d0a1a0a
+    ) {
+      throw new Error("not a PNG");
+    }
+    return { width: buf.readUInt32BE(16), height: buf.readUInt32BE(20) };
+  } finally {
+    await fh.close();
+  }
+}
 
 interface Props {
   /** Chronological list (oldest -> newest). */
@@ -89,9 +111,7 @@ export function ScreenshotModal({ screenshots, initialIndex, onClose }: Props) {
     const id = nextImageId();
     (async () => {
       try {
-        const meta = await sharp(path).metadata();
-        const w = meta.width ?? 1;
-        const h = meta.height ?? 1;
+        const { width: w, height: h } = await readPngDimensions(path);
         // Available cell box inside the chat region (reserve header /
         // footer so they stay visible). Cells are ~1:2 on screen, so
         // image cell-aspect = (h/w) / 2.


joshkotrous:
@cursor review this pr

jorgeraad (approved):
@maxgenovesi took a look.

Verified the fix on darwin-arm64:
- Pre-fix: launching the TUI crashes immediately with Could not load the "sharp" module using the darwin-arm64 runtime
- Post-fix: same launch boots the TUI clean. The @img/sharp-* tree is fully gone from the binary (sharp error string: 1 => 0, bundled modules 1598 → 1558).
  
LGTM! Thanks for the contribution!

PR #845, opened 2026-06-17
fix(tui): bump @opentui to 0.1.107 to fix startup crash

### LINKED ISSUE #844: TUI crashes on startup: RGBA.defaultBackground is not a function
The TUI fails to launch on `canary`:

```
TypeError: RGBA.defaultBackground is not a function
  at main (src/tui/index.tsx:697:27)
```

**Cause:** `createCliRenderer()` calls `RGBA.defaultBackground()` (added in #841), but that static method was only introduced in `@opentui/core@0.1.107`. `package.json` pins `^0.1.80`, so the lockfile had frozen `0.1.100` — which predates the method.

**Fix:** bump `@opentui/core`/`@opentui/react` floor to `^0.1.107` (patch bump, same `0.1.x` line — no breaking changes).

**Note:** `bun run build` smoke-tests the build but never calls `createCliRenderer`, so a renderer-init crash passes CI. A boot smoke test would catch this class of failure (follow-up).

jorgeraad:
Fixes #844.

## What
The TUI crashed on launch with `RGBA.defaultBackground is not a function` (src/tui/index.tsx:697). That static method, called by `createCliRenderer()` since #841, was only added in `@opentui/core@0.1.107` — but the loose `^0.1.80` floor let the lockfile pin the older `0.1.100`.

## Change
- `@opentui/core` and `@opentui/react`: `^0.1.80` → `^0.1.107` (patch bump within the same `0.1.x` line — no breaking changes)
- `bun.lock` resolves both to `0.1.107`

Raising the floor (not just the lock) prevents a fresh install from resolving back below the method's introduction.

## Verification
- `RGBA.defaultBackground()` returns a valid RGBA
- TUI boots to the splash, empty stderr (no crash)
- `tsc` clean (0 errors)

---

> [!NOTE]
> **Low Risk**
> Dependency-only patch bump within 0.1.x; no application code changes, fixes TUI launch regression.
> 
> **Overview**
> Raises the minimum `@opentui/core` and `@opentui/react` versions from **^0.1.80** to **^0.1.107** in `package.json`, with `bun.lock` pinning both to **0.1.107**.
> 
> This aligns installed OpenTUI with TUI startup code that calls `RGBA.defaultBackground()` during `createCliRenderer()` (e.g. in `src/tui/index.tsx`). That API exists only from **0.1.107** onward; a loose **^0.1.80** floor allowed resolves like **0.1.100** and caused launch failures (`RGBA.defaultBackground is not a function`). Bumping the semver floor—not only the lockfile—stops fresh installs from resolving below the required release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 751fa4d5da0a28f1e8243876c358daed6998db5c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kylejryan (approved):
LGTM!

PR #667, opened 2026-04-29
fix(offsec-agent): rename response tool input field to avoid name collision

jorgeraad:
Renames the `response` tool's inner input field from `response` → `result` so it no longer collides with the tool's own name. The collision was making Haiku 4.5 occasionally emit `{"response":{"response":{...}}}` instead of `{"response":{...}}`, which Zod rejected after the model had already streamed the full ~50KB tool input. Callers pass the schema, not the wrapper, so no other code changes.

Closes #666

---

> [!NOTE]
> **Low Risk**
> Small, localized API-shape change to a single tool input; main risk is breaking any external prompts/tests that still send `response` instead of `result`.
> 
> **Overview**
> Updates the offsec agent’s `response` tool contract so the tool input object uses `result` instead of `response`, and wires `execute` to pass `result` through to `onResult`.
> 
> This avoids occasional name-collision behavior where models emit an extra `response.response` wrapper that fails Zod validation, without changing how callers provide the response schema or how the captured result is consumed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c17ea23b6ffcb601fc5b630c91e8dc83848243e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #863, opened 2026-06-22
feat(pentest): generate next-run objectives at end of each run

Yuvanesh-ux:
## Summary

Makes subsequent pentests of the same endpoint productive instead of re-testing already-completed objectives. At the end of a run, the pentest orchestrator now spawns a focused planning sub-agent that reviews what was tested and proposes fresh objectives for the next run.

- New `generate_next_objectives` tool + `runObjectiveGenerationAgent` planning sub-agent (reasoning + `web_search`/`get_page` only), wired to the event bus as a child so it surfaces as its own timeline. Modeled on `threatModelGenerator`.
- New optional `newObjectives: string[]` field on `PentestResponseSchema` and `PentestResult`; `runTargetedPentestAgent` now returns it. Optional so existing console builds keep working.
- Orchestrator system prompt + instructions: new terminal "GENERATE NEXT OBJECTIVES" step, plus a zero-open-objectives path (skip worker fan-out, do light recon, then generate) so a would-be-wasted run seeds the next one.
- **Incidental fix:** `resolveResult` read the stale `response` key, but the response tool input field was renamed to `result` (#667). This made `objectiveResults` silently `undefined`. Now reads `result` and captures both `objectiveResults` and `newObjectives`.

## Consumer side

The console side (objective split + merge persistence) lands in a separate console PR that bumps the `packages/apex` submodule pointer to this branch.

## Test plan

- [x] `bun run tsc` — no new errors (pre-existing TUI/React-19 type noise only)
- [x] `bun run test src/core/agents/specialized/pentest/agent.test.ts` — 10 pass (incl. new newObjectives schema + orchestrator-prompt + zero-open assertions)
- [x] `bun run test src/core/agents/offSecAgent` — 232 pass (tool registration intact)
- [x] biome clean on changed files


joshkotrous:
I don't think this needs a new tool + sub-agent. The orchestrator already compiles `objectiveResults` from what each worker reported and has all the context it needs (coverage, outcomes, recon anomalies). We could just lean on the `newObjectives` field already on the response schema and have the orchestrator populate it directly as part of FINISH — same way it does `objectiveResults`. The schema description can carry the dedup/grounding guidance that's currently in the planning prompt.

That drops `generateNextObjectives.ts` entirely, avoids the extra sub-agent spawn/cost/latency, and removes the "copy verbatim from tool into response" round-trip. Keep the schema change and the `resolveResult` fix — those are the valuable bits.

joshkotrous (approved):

Yuvanesh-ux:
@cursoragent babysit untill ci checks pass

Yuvanesh-ux:
@cursoragent babysit the pr untill all ci checks pass

PR #908, opened 2026-07-10
feat(threat-model): enumerate framework-derived attack surface

jorgeraad:
The per-endpoint threat-model generator produces the threat model and pentest objectives for every endpoint — both whitebox and blackbox go through `document_endpoint` — but it never told the analysis agent to reason about framework-specific surface. So a Next.js page wouldn't get objectives for its server actions or RSC data fetches: surfaces reachable through the page that never show up as explicit routes.

This adds a step to the threat-model prompt that has the agent identify the framework (it already reads the source) and enumerate the implicit surface that framework implies, folding it into the endpoint's threat model and test plan. Next.js server actions are the one worked example, with an explicit instruction to generalize rather than pattern-match a fixed list of frameworks.

Only affects future recon runs — already-documented endpoints pick it up on re-recon.

---

> [!NOTE]
> **Low Risk**
> Prompt-only change to `threatModelGenerator.ts` with no runtime or auth logic changes; effect is richer (possibly longer) model output on future recon runs.
> 
> **Overview**
> Extends the per-endpoint threat-model generator so the analysis agent treats **implicit, framework-provided entry points** (not just explicit handler inputs) as in-scope for business logic, vectors, and pentest objectives.
> 
> **Part 1 (data flow)** now asks for inputs reachable through the endpoint—including channels like `server-action` / `rsc`—and calls out examples such as Next.js server actions and RSC fetches.
> 
> **Part 2** adds **Step 1b — recall framework-derived surface**: detect stack from code or description, then use a recall table of sharp-edge *classes* (hidden RPC/data URLs, debug pages, actuators, signing secrets, public env leakage, mass assignment, built-in proxies, deserialization) with example frameworks and probe signals. Instructions stress generalizing beyond the table, anchoring claims in evidence, and folding findings into Part 4’s test plan.
> 
> The Zod field descriptions for `businessLogic` and `threatModel` are aligned with the same wording so structured output matches the prompt.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1d0da8bcf6bbdf896dbbcf922e469a062a6546f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Yuvanesh-ux:
@jorgeraad I like where this is going, I think there might be an oppurtunity here to soup this even more. A "manual" or more framework specific insturctions would go a long way i believe. i.e specific weakness of next.js, springboot, svelte, etc.

jorgeraad:
@Yuvanesh-ux good point. Gave some examples of what to look for in the prompt to guide the agent without overloading the prompt or restricting the agent to a predetermined enumerated list.

If we find the agent is missing things (just like we found it wasn't considering server actions for Next.js before this change), we can tweak the prompt some more or even conditionally append a registry of things that tend to get missed.

kylejryan (approved):

PR #909, opened 2026-07-13
feat(fast-strike): bank findings mid-chain + target-aware exploit guidance

kylejryan:
## Summary
- Tighten fast-strike system prompt so confirmed vulns are documented immediately via `document_vulnerability` (including mid multi-step chains), not held until the final objective.
- Prefer target-specific exploit angles (endpoint/type/assembly/domain signals) over grinding generic playbook variants when a confirmed class resists the textbook exploit.
- Prompt-only change — no swarm-path or harvest-pipeline changes.

## Test plan
- [ ] Run a fast-strike session against a multi-step CTF/chain target and confirm intermediate findings land in the registry before the final flag/objective.
- [ ] Confirm swarm/full pentest path is unchanged (`fastStrike` flag still gates `runFastStrike` only).
- [ ] Spot-check that the updated prompt still emphasizes prune/speed and does not encourage wordlist grinding.


jorgeraad (approved):
LGTM

PR #910, opened 2026-07-13
fix(agents): protect verified creds + non-destructive proof-of-impact

jorgeraad:
Two related guardrails on the pentest/auth agents, prompted by a run that bricked auth on a shared test environment: it disabled team-wide MFA via a broken-authorization bug and overwrote a shared account's TOTP device to prove vulns, then crashed before any cleanup — locking the account out and leaving MFA off for everyone else on that environment.

**Never modify verified/provided credentials.** The auth sub-agent (shared by the login and credential-test paths) now treats credentials it was handed as already-verified and shared: never change the password, complete a reset / forced-password-change / account-recovery flow, or modify MFA/2FA settings for that account. When they don't authenticate (rejected, locked out, forced change, MFA required, email unconfirmed), fail fast — `complete_authentication` with success=false and a reason — instead of trying to recover. It also may not self-register a new account to work around a failed login (that account isn't the credential under test and pollutes results). REGISTER-mode tasks (no provided creds) are unaffected.

**Graded proof-of-impact ladder.** Replaces the blanket "stop at the point that proves impact" with a least-invasive-first ladder that keeps POCs verifiable while bounding blast radius:
1. Prove an authz / privileged-role / IDOR boundary with a read, or a benign reversible write to a low-impact field.
2. If a reversible state-changing write is the only convincing proof, capture the current value → change → capture evidence → restore, preferring your own / self-registered account. Do not rely on end-of-run cleanup alone; a crash can strip it.
3. Never complete an irreversible mutation against a shared/provided account (MFA/TOTP overwrite, single-use token consumption, email change, password rotation, deletion). Demonstrate reachability instead, or use an account you registered yourself.

Prompt-only; typechecks clean.

Not in this PR (tracked separately): a harness-level restore-journal + guaranteed teardown to make reversible-write cleanup survive crashes; environment/account isolation (the root cause — one run's mutations aren't isolated from others); and redaction of live tokens embedded in finding POCs.

---

> [!NOTE]
> **Low Risk**
> Changes are prompt and test text only—no runtime auth or tool behavior—so risk is limited to model compliance with new instructions.
> 
> **Overview**
> **Authentication agent** — When credentials are provided in the task prompt and in error-recovery guidance, the agent must treat them as **verified and shared**: no password changes, account-recovery flows, or MFA/2FA enrollment/disable/reset; on genuine login failure, **`complete_authentication` with `success=false`** instead of recovery; no self-registration as an auth workaround. Mechanical retries (field names, CSRF, endpoints) remain allowed.
> 
> **Pentest agent** — Replaces a single “don’t complete destructive account actions” note with a **graded proof-of-impact ladder**: prefer reads or benign reversible writes; if a reversible write is needed, change → evidence → **immediate restore** (don’t rely on end-of-run cleanup); **hard ban** on irreversible or security-weakening mutations (MFA/TOTP, reset tokens, email/password rotation, deletion) on shared/provided accounts, **other users via IDOR**, or **org-wide security settings**—prove authorization without completing the weakening step. Clarifies that **throwaway registration** is only for disposable POC targets, not as a substitute for logging in as the credential under test.
> 
> **Tests** — `agent.test.ts` expectations updated to assert the new pentest prompt strings (`Build verifiable POCs`, MFA/TOTP wording).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e22703a441491543f1e996f97b912ded69fa7955. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Yuvanesh-ux:
@jorgeraad pls address bugbot finding and I can take another look

jorgeraad:
Addressed the Bugbot finding in 9a124082. Removed the "self-created accounts are unrestricted" carve-out from the auth sub-agent and mirrored the pentest agent's ban: it may not self-register a new account to work around a failed provided credential — it fails fast with `complete_authentication` success=false instead. REGISTER-mode tasks are unaffected (they have no provided creds). The throwaway-account allowance stays where it belongs, in the pentest agent's proof-of-impact ladder.

jorgeraad:
Addressed the three latest Bugbot findings in 260eb788:

- **Ungated shared-credential system prompt** — gated the rules in `AUTH_SUBAGENT_SYSTEM_PROMPT` on *"if credentials were provided"*, so the always-present system prompt no longer asserts creds were provided when the credential manager is empty (no more disagreement with the no-credential user prompt).
- **Conflicting registration guidance** — scoped the `do NOT register` ban to *authenticating*, and made explicit that registering a throwaway account is allowed only as a disposable *target* for destructive-flow POCs (per the blast-radius rungs); aligned the rung wording to match.
- **Stale test contract** — updated `agent.test.ts` to assert the new proof-of-impact ladder strings instead of the removed destructive-action wording.

tsc + the pentest contract test pass locally.

jorgeraad:
@Yuvanesh-ux sure, addressed.

kylejryan (approved):

PR #914, opened 2026-07-15
feat: expose abortAndDrain for bounded Camoufox teardown

Yuvanesh-ux:
## Summary
- Add `startTargetedPentestAgent` handle with `abortAndDrain` so hosts can await Camoufox reap after timeout/pause
- Disconnect owned browser on abort; PID-starttime guards before descendant SIGKILL
- Keep `runTargetedPentestAgent` as a thin wrapper for existing callers

## Test plan
- [x] `targetedPentest.test.ts` covers handle + abortAndDrain
- [ ] Console blackbox holds endpoint slots until drain completes


jorgeraad (approved):
LGTM

PR #941, opened 2026-08-03
Durable-runtime seams: injectable subagent spawner and AI hooks

jorgeraad:
## What this does

Durable agent runs (crash-safe, exactly-once) need the agent runtime to be **injectable**: the points that spawn a child, mint an id, record usage, or call the model must be swappable without changing call sites. This PR adds those seams and wires **today's behavior as the default**, so nothing changes at runtime until a host injects a durable implementation, and nothing in apex does. It is the upstream prerequisite for the Console durable runtime (pensarai/console#2323).

## Size

Most of the line count is tests, so the total overstates the review. The behavior surface to review is the source diff.

| | Added | Removed | Files |
|---|---|---|---|
| Source | 1,066 | 821 | 32 |
| Tests | 397 | 1 | 15 |

## What changed

**1. One injectable subagent spawner.** Six tools (`spawn_pentest_agent`, `spawn_coding_agent`, `delegate_to_auth_subagent`, `document_vulnerability`, the attack-surface runner, and the threat-model fan-out) each hand-rolled the same child-spawn dance: a new session id, a new child event bus attached to the parent, the spawn and complete lifecycle, a bounded drain, and the same steps again on the error path. That is now one `SubagentSpawner` (`spawn` / `spawnMany`). The default `inProcessSubagentSpawner` is a faithful extraction that owns the whole lifecycle in one place, including the child event bus (created and attached here, not in each tool) and a guaranteed failed-status emit on every throw. A durable runtime swaps in an implementation that runs each child as a durable workflow, with no tool change.

**2. Three optional durability hooks.** Each defaults to exactly today's behavior; a durable runtime injects the alternate so a replayed run is deterministic and does not repeat side effects.

| Hook | Default | Durable runtime injects |
|---|---|---|
| `streamIdFactory` | random message and part ids | ids derived from stream position, so a replay reuses the same rows instead of duplicating the transcript |
| `usageRecorder` | the process-global usage callback | a per-run recorder, so usage is attributed per run and not double-counted on replay |
| `languageModelMiddleware` | the raw model | middleware that records each model call and replays it from the log instead of calling the model again |

All three thread from the agent input through the `ToolContext` and `SpawnRuntime` into `streamResponse`, and are inherited by every spawned child.

**3. Shared inputs on specialized agents.** `SpecializedAgentInput` gains `sandbox`, `extraTools`, and the three hooks. Each specialized agent extends `OffensiveSecurityAgent` and forwards `...base`, so these reach the base agent in one spread instead of being re-listed per constructor. Class names and inputs are unchanged, so every call site is untouched. `spawn_coding_agent` folds a durable code-agent child's returned attack surface into the parent registry; in-process children still return void.

## The theme

One owner per cross-cutting concern. A tool says *what* to spawn; the spawner owns *how* (the id, the child bus, the lifecycle, the drain). An agent lists its own config; `...base` carries the shared hooks and sandbox. Call sites stay declarative, there is one place to change the wiring, and one place for a durable runtime to plug in.

## Notes for reviewers

- **Session ids:** child ids run through the spawner's own id factory and the root comes from the dispatched request, so both are covered. A few direct `newSessionId()` calls (the endpoint-documentation agent and the whitebox workflow) still bypass any seam. They are fine today and only matter once those flows run durably.
- **Usage and cache:** the cached-versus-uncached split reaches a durable consumer on `onCacheMetrics`, separate from the per-run totals. A durable runtime that bills usage must consume both callbacks and join them per step.

## Compatibility and verification

Additive: every new input is optional, and unset preserves current behavior. `lint`, `tsc`, `test`, `build`, and `knip` pass.

No linked issue. Upstream prerequisite for pensarai/console#2323.

---

> [!NOTE]
> **Medium Risk**
> Touches core agent orchestration, subagent lifecycle, and streaming id minting across many tools; behavior is intended to be unchanged when hooks are unset, but regressions in nested spawns or event nesting would affect scans and UI.
> 
> **Overview**
> Introduces **injectable seams** so a durable host can swap how children are spawned and how model calls/usage/ids are handled, while **defaults preserve today’s in-process behavior**.
> 
> A new **`SubagentSpawner`** (`spawn` / `spawnMany`) replaces duplicated child wiring in orchestration tools (pentest workers, coding agents, auth delegation, finding judge, attack surface, threat-model fan-out). **`inProcessSubagentSpawner`** owns id minting, child event bus attachment, lifecycle events, bounded drain, and failed completion on errors.
> 
> **`OffensiveSecurityAgent`** and **`streamResponse`** accept optional **`languageModelMiddleware`**, **`usageRecorder`** (per-run usage instead of the global callback), and **`streamIdFactory`** (deterministic message/text/tool-part ids from step indices). These flow through **`ToolContext`** / **`SpawnRuntime`** into every spawned child. **`runWithAiPayloadCapture`** gates OTEL payload recording per async run.
> 
> **`SpecializedAgentInput`** centralizes **`sandbox`**, **`extraTools`**, spawner, and hooks; specialized agents forward via **`...base`**. Pentest swarm and whitebox workflows use **`spawner.spawnMany`** and forward the durable hooks to workers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52a80a8bd975841e7c4ada4a9ff0dca3846089c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jorgeraad:
All checks are green and the branch is rebased onto current canary. This is now intended to merge independently before the pointer-only Console precursor. One write-access approval is the remaining repository-rule requirement; review requested from @kylejryan and @Yuvanesh-ux.

Yuvanesh-ux:
@jorgeraad pls resolve bugbot dinings and I can review

jorgeraad:
@Yuvanesh-ux all good now

Yuvanesh-ux (approved):

PR #951, opened 2026-08-13
feat(ai): report the prompt-cache token split through the usage callback

jorgeraad:
Prompt caching works (#943), but the token split it produces is discarded at the usage-reporting boundary. `UsageCallback` forwards only `(model, inputTokens, outputTokens, ctx)`, so a consumer can see that a call happened but not how much of it was served from cache.

That matters because the two halves are priced very differently — providers bill a cache read at roughly a tenth of the input rate and a cache write at a premium — so a consumer that can't see the split has to price every token as if it were uncached.

## What changed

- `UsageCacheDetails { cacheReadTokens, cacheWriteTokens }` is exported.
- `UsageCallback` gains an **optional trailing** `cache?: UsageCacheDetails`. Optional by design: every existing consumer keeps compiling untouched.
- `getUsageCacheDetails(usage, providerMetadata)` resolves the split, preferring the AI SDK's normalized `usage.inputTokenDetails` and falling back to `providerMetadata.anthropic`, then zero.
- Both `_usageCallback` call sites pass it — the `onStepFinish` wrapper in `streamResponse`, and `generateObjectResponse` (which now also destructures `providerMetadata` from its `generateText` call).

## One thing worth knowing about the numbers

`usage.inputTokens` **includes** the cached tokens; it is not the uncached remainder. Measured against live Bedrock on Claude Haiku 4.5:

```
call 1 (cache write):  inputTokens=16220  details={noCacheTokens:13, cacheRead:0,     cacheWrite:16207}
call 2 (cache read):   inputTokens=16220  details={noCacheTokens:13, cacheRead:16207, cacheWrite:0}
                       totalTokens=16224 = inputTokens + outputTokens
```

`13 + 16207 = 16220`. A consumer that prices `inputTokens` at the full rate *and* prices `cacheReadTokens` separately will therefore count the cached portion twice. The normalized shape also exposes `noCacheTokens`, which is the uncached remainder directly.

Reading the provider source alone suggests the opposite conclusion, so this was settled by measurement rather than inference.

## Not changed

`onCacheMetrics` has its own extraction with the **opposite** precedence — `providerMetadata.anthropic` first, normalized second — because the gateway provider does not populate `providerMetadata.anthropic`. The two agree whenever only one source is populated, which is every case observed. Folding them together would change existing metric behaviour, so it is left alone; worth revisiting separately.

## Verification

`bun run tsc` clean. Full suite: 1520 passed, 17 skipped. New unit tests cover the normalized path, the provider-metadata fallback, and absent usage returning zeros.

---

> [!NOTE]
> **Low Risk**
> Additive, backward-compatible API change to usage reporting with focused tests; no auth or runtime behavior changes beyond extra callback arguments.
> 
> **Overview**
> **Usage reporting** now forwards prompt-cache splits so billing consumers can price cache reads and writes separately instead of treating all input tokens as uncached.
> 
> The global `UsageCallback` gains an optional trailing `cache?: UsageCacheDetails` (`cacheReadTokens`, `cacheWriteTokens`). New helper `getUsageCacheDetails` resolves those counts from the AI SDK’s `usage.inputTokenDetails`, with Anthropic `providerMetadata` as fallback. Both reporting paths—streaming `onStepFinish` in `streamResponse` and `generateObjectResponse` (which now reads `providerMetadata` from `generateText`)—pass the cache details into `_usageCallback`.
> 
> Unit tests cover normalized usage, metadata fallback/precedence, and zero defaults when usage is missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36e1199dc5d54fa59eb12f43ea584b4b54776d4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jorgeraad:
Implemented elsewhere. Tracked by #2461 and #2658

Issue #1013, opened 2026-08-28
Emit complete inference attempt lifecycles and align local usage projections


Yuvanesh-ux:
## Problem

Apex's local usage views, session metrics, traces, and telemetry do not all observe the same inference operations or terminal states. Some helper calls and interrupted paths can bypass a usage projection, while root/subagent ownership uses different routing paths.

After a provider-attempt contract exists, every inference operation should emit the same lifecycle events and local projections should consume those events consistently while remaining explicitly best-effort and non-financial.

## Scope

- Emit attempt lifecycle events for agent streams, nested agents, structured generation, summaries, tool repair, session naming, CVSS, finding helpers, and benchmark/direct paths.
- Record started, completed, failed, partial, aborted, and retry outcomes according to the provider-attempt contract.
- Ensure an observer failure cannot suppress later observers or cause a successful provider request to be repeated.
- Carry stable root/child session, run, logical-step, retry-parent, and attempt attribution through workflows and resume paths.
- Derive `execution-metrics.json`, JSONL traces, W&B, and OTLP from canonical attempt events where applicable.
- Prevent late events from writing to a newly active session and prevent root/subagent double counting.
- Preserve the explicitly best-effort and non-financial nature of local files and observability exporters.

## Acceptance criteria

- [ ] Every supported inference entry point has a documented attempt disposition, including host-owned and non-billable paths.
- [ ] Completed, failure, retry, partial, and abort fixtures produce the expected lifecycle sequence without duplicate attempts.
- [ ] Callback ordering/failure tests prove that an observer cannot retry or hide an already-successful model request.
- [ ] Nested agents and resumed workflows retain stable attribution without using display labels or trace indexes as financial identities.
- [ ] Session metrics, JSONL, W&B, and OTLP projections have clear source data and do not double count the same attempt.
- [ ] Projection write/export failure does not change the underlying attempt outcome.

## Non-goals

- No pricing, credit handling, provider settlement, customer identifiers, or external billing behavior.
- No raw prompts, responses, target data, credentials, or provider payloads in attempt events.

## Related public work

- [#903](https://github.com/pensarai/apex/issues/903) owns the remaining slug-as-subagent-ID migration and should be coordinated, not duplicated.
- [#996](https://github.com/pensarai/apex/pull/996) through [#1008](https://github.com/pensarai/apex/pull/1008) introduced the current session usage and telemetry projections.

PR #1021, opened 2026-08-31
Make token usage reports honest and consistent

Yuvanesh-ux:
**TLDR**
- Token counts from different AI providers did not mean the same thing
- Missing cache data was treated as zero, so “no cache” and “we don’t know” looked identical
- This PR adds one shared usage record per model call, keeps unknown values as unknown, and leaves existing dashboards working
- Not wired into live model calls yet — that is a follow-up (#1013)

**What’s the problem?**

Apex already reports how many tokens a model used, but the numbers come in different shapes. Anthropic’s `input_tokens` is uncached. The AI SDK’s `inputTokens` already includes cache. Some routes omit cache fields entirely.

The current callbacks fill in missing cache counts with `0`. A downstream consumer cannot tell a real zero from “the provider never sent that field,” and they cannot assume every input total has the same cache meaning.

**How we fixed it**

We added a versioned `InferenceAttempt`: one record for one real model request (`src/core/ai/inference-attempt`).

- Give the request an id before it is sent. Retries get a new attempt id and keep the same idempotency key
- Normalize tokens into inclusive input, uncached input, cache read, cache write, and output
- Keep missing provider data as `null` instead of pretending it is `0`
- Reject counts that contradict each other (inclusive must equal uncached + cache read + cache write when all four are known)
- Cover each provider transport with fixtures
- Keep today’s `onUsage` / `onCacheMetrics` / `onStepFinish` callbacks as they are, with a small helper to map the new record onto that older shape

No pricing, billing, or customer fields. This module is not used in `streamResponse` yet.

**Built on**

This sits on top of recent usage work: [#1002](https://github.com/pensarai/apex/pull/1002) added cache-read/write counts to the usage callback, and [#943](https://github.com/pensarai/apex/pull/943) made prompt caching actually fire on Bedrock and the Pensar gateway. The session usage footer and metrics PRs ahead of those (#996–#1000) are what those callbacks feed today.

<div><a href="https://cursor.com/agents/bc-7ae2d25e-fe2d-43d6-b12d-5ab34f8cecc7?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7ae2d25e-fe2d-43d6-b12d-5ab34f8cecc7&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Yuvanesh-ux:
@cursor take a look at the bugbot findings in this pr and resolve them as well as rebasing with latest staging

src/core/attempt/builder.ts
```diff
@@ -0,0 +1,150 @@
...
+  return {
+    attemptId: identity.attemptId,
+    idempotencyKey: identity.idempotencyKey,
+    started,
+    complete: (usage) => settle("completed", usage),
+    fail: (usage) => settle("failed", usage),
+    abort: (usage) => settle("aborted", usage),
+    partial: (usage) => settle("partial", usage),
+    retried: (usage) => settle("retried", usage),
+    retry: () =>
```
jorgeraad: Check this from Claude:

`retry()` mints a fresh `idempotencyKey` for every physical attempt, and
`attempt.test.ts:123` asserts that it differs from the parent's.

An idempotency key that changes on every retry cannot deduplicate anything, which is the
one job it has. As written it is a second unique id for the same object, from the same
generator, in the same shape, differing only in prefix.

Two ways out. Make it stable across a retry lineage, minted at the root and carried
through `retry()`, which also gives it a reason to exist alongside `attemptId`. Or drop
the field. Cheap to settle now, much harder once #1013 emits it and something downstream
keys on it.

Smaller thing in the same function: calling `retry()` twice on one handle produces two
attempts both at `sequence: 2` with the same `previousAttemptId`, because `lineage` is
captured in the closure. Chains work fine, fan-out silently duplicates a sequence number.
Either document `retry()` as single use or derive the sequence from the previous handle.

While we are here, `idk_` reads as "I don't know" at a glance. If the field goes away,
so does that.

jorgeraad (dismissed):
Was hard for me to infer based on the file and class names. You mind renaming:
- src/core/attempt =>  src/core/ai/inference-attempt/
- ProviderAttemptEnvelope => InferenceAttempt

Also want to call out this code isn't being used yet? That's good if we want to keep PRs small, but would be helpful in future to put in stack with PR that actually wires it in.

Yuvanesh-ux:
@jorgeraad pls check now

kylejryan (approved):

PR #1059, opened 2026-09-08
fix(observability): export open spans on shutdown

kylejryan:
### What does this PR do?

Apex previously ended only registered root agent spans during SIGTERM shutdown. OpenTelemetry does not export a span until `span.end()` invokes the processor chain, so any still-open generation, tool, or nested agent spans were absent even though the provider was force-flushed. That left consumers with orphaned children and forced EvalGate to infer missing parents.

This change:
- installs a lightweight `SpanProcessor` ahead of the exporter to track every recording span owned by Apex’s standalone SDK
- ends all still-open spans in reverse start order (children before parents) before the final force-flush
- marks shutdown-ended spans with ERROR status and `pensar.telemetry.interrupted=true`
- avoids double-ending spans still present in the legacy root registry
- preserves embedded-host isolation because the processor exists only on Apex’s standalone provider

- `bunx vitest run src/core/observability/shutdown.test.ts src/core/observability/runtime.test.ts src/core/observability/trace-contract.test.ts` (49 tests)
- Added an OTLP receiver integration test proving a still-open root/model/tool tree is exported with intact parent IDs during shutdown
- `bun run tsc`
- `bun run build`

---

> [!NOTE]
> **Medium Risk**
> Changes bounded process shutdown and span lifecycle for the standalone OTLP runtime; incorrect ordering could affect trace export or double-end behavior, though guards limit blast radius to telemetry.
> 
> **Overview**
> **Fixes missing trace data on Apex shutdown** by closing every still-open span before OTLP flush, not only registered root agent runs. OpenTelemetry only exports spans after `onEnd`, so previously nested model/tool spans could disappear while consumers saw orphaned children.
> 
> Adds an **`ActiveSpanProcessor`** as the first processor on Apex’s standalone tracer provider. It tracks recording spans (with a cap), ends them **deepest-first** during shutdown, and tags them with `pensar.telemetry.interrupted`, ERROR status, and `ApexProcessInterrupted` when appropriate. **`startObservabilityRuntime` shutdown** now runs `endAll()` before `forceFlush`, clears the legacy root registry, then runs **`endAll()` again** after flush to catch spans started while flush yields.
> 
> **Idempotent `span.end()`**: root registry, agent `invoke_agent` finally blocks, and shutdown paths all guard with **`span.isRecording()`** so shutdown-forced ends don’t race normal run completion.
> 
> Tests cover child-first interruption, late spans during flush, and OTLP export of an open root/model/tool tree with parent links intact.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f53b28b53296eaf3d9351c2afbf2ce91b8a8c933. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/core/agents/offSecAgent/offensiveSecurityAgent.ts
+++ b/src/core/agents/offSecAgent/offensiveSecurityAgent.ts
@@ -999,8 +999,9 @@ export class OffensiveSecurityAgent<TResult = void> {
           } finally {
             unregisterActiveRootSpan(span);
             // Ends only after runConsume's finalization (persistence +
-            // owned-resource disposal) has settled.
-            span.end();
+            // owned-resource disposal) has settled. Shutdown may have already
+            // ended it while that finalization was still in flight.
+            if (span.isRecording()) span.end();
           }
         },
       );
--- a/src/core/observability/active-root-spans.ts
+++ b/src/core/observability/active-root-spans.ts
@@ -21,7 +21,7 @@ export function unregisterActiveRootSpan(span: Span): void {
 export function endAllActiveRootSpans(): void {
   for (const span of activeRootSpans) {
     try {
-      span.end();
+      if (span.isRecording()) span.end();
     } catch {
       // A dead span must never block process teardown.
     }
--- a/src/core/observability/active-spans.ts
+++ b/src/core/observability/active-spans.ts
@@ -1,4 +1,4 @@
-import { SpanStatusCode } from "@opentelemetry/api";
+import { diag, SpanStatusCode } from "@opentelemetry/api";
 import type {
   ReadableSpan,
   Span as SdkSpan,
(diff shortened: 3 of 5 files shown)


kylejryan:
Follow-up hardening is in `f53b28b5` (review findings addressed): interrupted exports are asserted at the real OTLP/BatchSpanProcessor boundary, spans created during the final force-flush window are ended before downstream shutdown, active tracking is bounded and re-entrancy-safe, interruption gets a consistent `error.type`, and the agent finalizer avoids double-ending a span already closed by shutdown. The targeted shutdown suite now has 17 passing cases; typecheck and Biome are clean.

Yuvanesh-ux (approved):
Noticed this across my PRs as well as a whole I think the comments are too verbose, might be worth adding in the repo a short guide on writing code comments.

PR #1070, opened 2026-09-17
fix(cvss): require a schema-enforcing endpoint for GLM structured output

jorgeraad:
### What does this PR do?

The CVSS 4.0 scorer has failed on every finding in a GLM-5.2 pentest since 2026-09-09. When it fails, `document_vulnerability` substitutes a conservative fallback score of 5.0 / MEDIUM with a blank vector, and because the Console derives a finding's severity from its CVSS score, the fallback also sets the severity. Affected findings reach the customer as Medium / 5.0 with no vector. Telemetry counts 86 failures and 0 successes since the change landed; the 45 days before it were 134 failures against 2,429 successes.

Pinning GLM 5.2 to Z.ai added `provider: { only: ["z-ai"], allow_fallbacks: false }` to `generateObjectResponse` as well as `streamResponse`. Z.AI's OpenRouter endpoint is the only provider for `z-ai/glm-5.2` that does not advertise `structured_outputs`, and OpenRouter's `require_parameters` defaults to `false`, so `response_format` was only a soft preference. Nothing enforced the schema, the model returned text the schema rejected, and every attempt fell through to the fallback.

Three changes:

- Structured generation now routes with `order: ["z-ai"], allow_fallbacks: true, require_parameters: true`. It still prefers Z.ai, but requires an endpoint that enforces the schema rather than silently downgrading. Streaming keeps the hard pin, since removing it altogether restores the intermittent failure rate the pin was added to fix.
- The fallback is logged at `warn` with the real error. It previously reached only the finding markdown and the tool-result string, and both failed model calls recorded as successful generation spans (the HTTP call succeeds; the parse throws), which is why this ran for nine days without anyone noticing.
- The second scoring attempt is gone. It re-sent the same prompt to the same model two seconds later and has never recovered a call.

The persisted finding already carries `cvss.scored: false` for this case. The Console half, which plumbs that flag into the issue record so a fallback score is distinguishable from a real one, is pensarai/console#2879. The two are independent and can merge in either order.

Reported in pensarai/console#2876.

No live model calls — the provider options are asserted against a mock model, which is the whole surface that changed.

- `bun run test` — 2772 passed, 18 skipped
- `bun run tsc`
- `bun run lint`
- `bun run build`

New coverage:

- `src/core/observability/telemetry.test.ts` — GLM structured generation sends the schema-enforcing routing; GLM streaming still sends the hard pin.
- `src/core/ai/ai.test.ts` — unit tests for both provider-option builders.
- `src/core/agents/offSecAgent/tools/documentFinding.test.ts` — a failed scorer marks the finding `scored: false` and warns with the real error, calls the scorer exactly once (the retry stays gone), and a successful scorer still marks the finding scored and logs nothing.

No screenshots — no user-facing surface changed.

### A note on the red check

`console-typecheck` is failing, and it fails for every apex PR opened against `canary` right now, not because of this diff. Console `staging` pins apex `dd54cc53`, which lives on `feat/finding-credential-provenance` (#1068) and is not on `canary` yet, so typechecking Console against `canary` reports `credentialIds` missing from `Finding` and `id` missing from the credential type — neither symbol appears anywhere in this diff. It last passed on `canary` at 2026-09-15 19:39Z, before those two Console PRs landed on 2026-09-16 19:51Z. It should go green once #1068 merges. Every other check is green.

---

> [!NOTE]
> **Medium Risk**
> Changes how CVSS and other structured LLM calls route on OpenRouter and alters failure observability; impact is limited to GLM models and scoring fallback behavior, not auth or data paths.
> 
> **Overview**
> Fixes GLM CVSS structured scoring by **splitting OpenRouter routing**: streaming/tool paths keep the hard **Z.ai-only** pin, while **`generateObjectResponse`** (CVSS and other schema-bound calls) now prefers Z.ai but sets **`require_parameters: true`** with fallbacks so OpenRouter picks an endpoint that actually enforces `response_format`.
> 
> **`document_vulnerability`** drops the second CVSS attempt and the 2s retry loop; on failure it still applies the MEDIUM fallback but emits a **structured WARN** (real error, cancelled flag, session/finding context) while persisted findings stay **`cvss.scored: false`**.
> 
> Tests lock in the two provider-option shapes, telemetry for GLM stream vs structured generate, and CVSS fallback logging / single-call behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33c3043993232e8f8d9c07be64761bc6c497256a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jorgeraad:
bugbot run

kylejryan (approved):

